PR #14, opened 2011-09-15
Create directories and symlinks

genadyp:


kolmanv:
Please see one comet.
Everything else is ok.
Please add Indexer and Crawler to fileutil.

PR #17, opened 2012-04-09
Request to pull Logger gem

yarondbb:


kolmanv:
Please see comments inside the commit.
https://github.com/yarondbb/bbfs/commit/778170b22742ddea0fff09c425cc389e31276264

yarondbb:
I have made the Logger a module. This way it will used globaly
The usage is as simple as it gets:

require 'logger'
User_Mdoule
  Logger.logInfo("this is a test info message level 0")
  Logger.logError("this is a test error message level 0")
  Logger.logWarning("this is a test warning message level 0")
      Logger.logDebug1("this is a test debug 1 message level 0")
      Logger.logDebug2("this is a test debug 2 message level 0")
      Logger.logDebug3("this is a test debug 3 message level 0")
      #start write  using level 1
      PARAMS.logger_debug_level = 1
      Logger.logInfo("this is a test info message level 1")
      Logger.logError("this is a test error message level 1")
      Logger.logWarning("this is a test warning message level 1")
      Logger.logDebug1("this is a test debug 1 message level 1")
      Logger.logDebug2("this is a test debug 2 message level 1")
      Logger.logDebug3("this is a test debug 3 message level 1")
     ...
end

The output example for one line:
BBFS logger 2012-04-20 13:53:11 +0300[INFO]:logger_test.rb:11:this is a test info message level 0

Regarding saving the log data in memory rather than flush it to a log file on every line. What are the prons and Cons?
Cons: saving I\O execution time - this depends on the intensity of the loggs.. every uS\mS\S\1000S..
           File is not opened all the time but only on flushes. If system crushes, it makes it more likabke to recover the file
Prons: Data can explode the process memory space if not flushed. We can set a max of some MBtyes before flushing (also depends on system memory)

I still need to implement this part and to add tests

kolmanv:
Please add:
1) logger directory and add version.rb file there (see examples).
2) Add logger.gemspec file.

I think the name logger is already taken (gem search logger -r)
You can rename to logging or log see yourself.

You are correct with the prons/cons. There are few additional pros for memory:
1) Easier to test. In your case, linux system will have different CR character (new line) and might fail the test due to that.
2) In the near future we will have many many servers and we will want to log to remote computers. In that case we need memory storage anyway.

Code is clean and good.
After you add the missing files I will pull.

PR #19, opened 2012-04-12
Kolmanv

kolmanv:
Gena, please review code.
I have made all tests pass.


genadyp:
We are building few files and directories for test purposes in monitor_path_test (test_file.1500 etc.)
Do we need them in git?
If we don't need them then let's remove them from commit and add a teardown method that removes them in test after usage.
Also we can check usage of Tempfiles: http://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html

kolmanv:
See commit in which I fixed windows tests and time modification.

kolmanv:
removed.

On Sun, Apr 15, 2012 at 7:07 PM, genadyp <
reply@reply.github.com

PR #24, opened 2012-05-03
Log: Consumer Producer new design

yarondbb:


yarondbb:
Made all comments. The Semaphor is redundant. I removed it.

PR #31, opened 2012-05-14
Proj conf

yarondbb:
Params new capability to load from yml
Logger corrections (sorry mistakenly I have worked on this on the same branch..)

### DIFF
--- a/lib/params.rb
+++ b/lib/params.rb
@@ -3,24 +3,25 @@ require 'yaml'
 # A simple params module.
 module BBFS
   module Params
-    def parameter(name, default_value, description)
+    def Params.parameter(name, default_value, description)
       self.instance_variable_set '@' + name, default_value
       self.class.send('attr_accessor', name)
     end
 
-    def read_from_yml file_path
-      proj_params = YAML::load_file(file_path)
-      proj_params.keys.each do
-      |key|
-        if not self.instance_variable_get '@' + key
-           puts "loaded yml param:'#{key}' which does not exist in Params module."
+    # Load yml params and override default values.
+    # Return true, if all yml params loaded successfully.
+    # Return false, if a loaded yml param does not exist.
+    def Params.read_yml_params yml_params_string
+      proj_params = YAML::load(yml_params_string)
+      proj_params.keys.each do |param_name|
+        if not self.instance_variable_get '@' + param_name
+          puts "loaded yml param:'#{param_name}' which does not exist in Params module."
           return false
         end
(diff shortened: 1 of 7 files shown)


lib/params.rb
```diff
@@ -7,7 +7,20 @@ def parameter(name, default_value, description)
       self.instance_variable_set '@' + name, default_value
       self.class.send('attr_accessor', name)
     end
-    module_function :parameter
+
+    def read_from_yml file_path
+      proj_params = YAML::load_file(file_path)
+      proj_params.keys.each do
+      |key|
```
kolmanv: move |key| to previous line.
.each do |key|

test/log/log_test.rb
```diff
@@ -14,24 +14,23 @@ module Log
...
-          return @data_list
-      end
-      def init
-        @data_list.clear
+        @test_queue.push data
       end
     end
 
     class TestLog < Test::Unit::TestCase
+
```
kolmanv, line 30: Remove this line.

test/params/test_1.yml.bak
```diff
@@ -0,0 +1,7 @@
```
kolmanv: Remove this file from your commit.

test/params/params_test.rb
```diff
@@ -0,0 +1,42 @@
...
+module BBFS
+
+  class TestLog < Test::Unit::TestCase
+
+    def test_define_both_code_and_yml_params
+      # Test set phase
+      code_param_name =  'param_1'
+      code_param_value = 'value_1_code'
+      yml_param_value = 'value_1_yml'
+      yml_file_name = 'test/params/test_1.yml'
```
kolmanv: You don't have to work with files here.
You can inline write content and load yml from string.
This is better because the test is more dynamic, you don't have to edit few files if things change.
Better remove the yml test files (unless you really have a good reason to leave them) and test with strings.
This is always the preference.
Just do Yaml.load(yaml string)

kolmanv:
Yaron, please notice that you had to create two branches here.
I want to merge the log fixes you did well, but I cannot because there are not ready things in Params.
GIT enables working with many many branches in parallel. This is good because of concern separation.

kolmanv:
Ohhh, now I see your comment at the beginning. Disregard the previous comment.

kolmanv:
Yaron, did you fix this pull request? May I merge?

yarondbb:
Yes, I fixed the issues. you can merge
1 question: if I do not test load from file then how would I add a method to load from file?

PR #34, opened 2012-05-21
Fixed file generator utility after code review

slavapa:
Please review.

### DIFF
--- a/lib/file_utils/file_generator/file_generator.rb
+++ b/lib/file_utils/file_generator/file_generator.rb
@@ -1,13 +1,12 @@
 require 'digest'
 require 'file_utils/file_generator/file_generator_parameters'
 require 'fileutils'
-require 'params'
 
 #Generates files with random content with random file size
 # according to given FileGeneratorParameters
 class FileGenerator
   attr_accessor :file_gen_params
-  attr_reader :one_kb_string, :one_mb_string
+  attr_reader :one_mb_string
 
   def initialize (file_gen_params = FileGeneratorParameters.new())
     @file_gen_params = file_gen_params
@@ -15,10 +14,6 @@ class FileGenerator
 
   def one_mb_string
     @one_mb_string || @one_mb_string = prepare_one_mb_string
-  end
-
-  def one_kb_string
-    @one_kb_string || @one_kb_string = prepare_one_kb_string
   end
 
   def get_small_rand_unique_str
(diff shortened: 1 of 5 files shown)


lib/file_utils/file_generator/file_generator.rb
```diff
@@ -1,140 +1,122 @@
-require 'fileutils'
-require_relative 'file_generator_parameters'
 require 'digest'
+require 'file_utils/file_generator/file_generator_parameters'
+require 'fileutils'
+require 'params'
```
kolmanv: Why did you add the library if you are not using it?

lib/file_utils/file_generator/file_generator.rb
```diff
@@ -1,140 +1,122 @@
...
+class FileGenerator
+  attr_accessor :file_gen_params
+  attr_reader :one_kb_string, :one_mb_string
+
+  def initialize (file_gen_params = FileGeneratorParameters.new())
+    @file_gen_params = file_gen_params
+  end
+
+  def one_mb_string
+    @one_mb_string || @one_mb_string = prepare_one_mb_string
```
kolmanv, line 16: one_mb_string is same as one_kb string.
Make one method which gets a parameter.

lib/file_utils/file_generator/file_generator.rb
```diff
@@ -1,140 +1,122 @@
...
+    "#{Time.new.to_i}_#{get_small_rand_unique_str}"
+  end
+
+  def get_random_letter
+    (rand(122-97) + 97).chr
+  end
+
+  def determine_dir_recursively (path)
+    if File.directory?(path)
+      #puts "The given path already exists #{path}"
```
kolmanv: change the whole if to one line:
FileUtils.mkdir_p path unless File.directory?(path)

delete this function if you use it only once.

lib/file_utils/file_generator/file_generator.rb
slavapa: I thought to use.
I'll use this or delete this in the next commit today. 

-----Original Message-----

kolmanv: ok, use the same pull request (i.e. just commit to the branch).

On Tue, May 22, 2012 at 9:12 AM, Slava <
reply@reply.github.com

slavapa: Removed require 'params'

lib/file_utils/file_generator/file_generator.rb
slavapa: Fixed

lib/file_utils/file_generator/file_generator.rb
slavapa, line 16: I dropped one_kb string at all.

kolmanv:
You have not removed the parameters file.
Please talk with Yaron and fix this Issue.
I will merge but only because this code is already in!

PR #35, opened 2012-05-23
Fix log file name print

yarondbb:


lib/log.rb
```diff
@@ -81,7 +81,7 @@ def Log.add_consumer consumer
     # formatting the data and push to consumers
     def Log.basic msg, type
       Log.init if not @log_initialized
-       /([a-zA-Z0-9\-_\.]+\.rb:\d+)/ =~ caller[1]
+       /([a-zA-Z0-9\-_\.]+:\d+)/ =~ caller[1]
```
kolmanv, line 84: Why did tests did not catch this bug?
Consider adding a test (should be one liner).

yarondbb, line 84: Since it is not realy a bug but only a bad definition of a requirement that a file name should be a ruby file.
what I am realy worried is why the typo  in the lib/params.rb did not break the tests.. My tests were broken.. so why the tests in the BBFS master did not break ?

PR #40, opened 2012-05-29
Added handling to use Params.parameter instead of YAML file for FileGene...

slavapa:
...rator.


kolmanv:
Yaron, please handle this. Slava this is his first review so patience :)

PR #58, opened 2012-06-16
Added require log

yarondbb:
sorted require lines according to style guid

### DIFF
--- a/lib/content_server.rb
+++ b/lib/content_server.rb
@@ -1,13 +1,16 @@
-require 'content_data'
-require 'file_copy'
-require 'file_indexing'
-require 'file_monitoring'
-require 'params'
 require 'set'
 require 'thread'
 
+require 'content_data'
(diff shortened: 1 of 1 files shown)


lib/content_server.rb
```diff
@@ -1,14 +1,17 @@
...
 require 'file_monitoring'
+require 'log'
 require 'params'
-require 'set'
-require 'thread'
-
-require 'content_server/content_receiver'
 require 'content_server/queue_indexer'
 
+
```
kolmanv, line 13: Are those extra lines needed by the new style guide?

kolmanv:
Please merge the changes, see tests pass.

lib/content_server.rb
yarondbb, line 13: No. They are dedundant.
Also, the author name is missing.. Pls add it later

PR #60, opened 2012-06-25
Added require 'log' for all files who use the log module to avoid 'undef...

yarondbb:
...ined' errors


kolmanv:
Did you add 'log' dependencies to the gemfiles?

yarondbb:
No, forgot about this demend. Will do it today.

PR #62, opened 2012-06-29
Added init sequence

yarondbb:
Added init sequence
Added rspec tests for Params module
Modified tests in params_test.rb to allign with changes

### DIFF
--- a/lib/params.rb
+++ b/lib/params.rb
@@ -1,45 +1,203 @@
+# Author: Yaron Dror (yaron.dror.bb@gmail.com)
+# Description: The file contains 'Params' module implementation.
+# Notes:
+#   module.init should be called if user wants to override defined parameters by file or command line.
+#   Parameter can be defined only once.
+#   Parameters have to be defined before they are accessed. see examples below.
+#
+# Examples of definitions:
+#   Params.string 'parameter_str', 'this is a string' ,'description_for_string'
+#   Params.integer 'parameter_int',1 , 'description_for_integer'
+#   Params.float 'parameter_float',2.6 , 'description_for_float'
+#   Params.boolean 'parameter_true', true, 'description_for_true'
+#   Params.boolean 'parameter_false',false , 'description_for_false'
+#   Note. Parameters have to be defined before they are accessed.
+#
+# Examples of usages (get\set access is through the [] operator).
+#   local_var_of_type_string = Params['parameter_str']
+#   puts local_var_of_type_string  # Will produce --> this is a string
+#   Params['parameter_float'] = 3.8
+#   puts Params['parameter_float']  # Will produce --> 3.8
+#   Params['parameter_float'] = 3
+#   puts Params['parameter_float']  # Will produce --> 3.0 (note the casting)
+#   Note. only after definition, the [] operator will work. Else an error is raised.
+#     Params['new_param'] = true  # will raise an error, since the param was not defined yet.
+#     Params.boolean 'new_param', true, 'this is correct'  # this is the definition.
+#     Params['new_param'] = false  # Will set new value.
+#     puts Params['new_param']  # Will produce --> false
+#
+# Type check.
+#   A type check is forced both when overriding with file or command line and in [] usage.
+#   if parameter was defined as a Float using Params.float method (e.g. 2.6) and user
+#   provided an integer when overriding or when using [] operator then the integer will
+#   be casted to float.
+#
+# Params.Init - override parameters.
+#   implementation of the init sequence allows the override of
+#   defined parameters with new values through input file and\or command line args.
+#   Note that only defined parameters can be overridden. If new parameter is parsed
+#   through file\command line, then an error will be raised.
+#   Input config file override:
+#     Default configuration input file path is: '~/.bbfs/conf/<executable name>.conf'
+#     This path can be overridden by the command line arguments (see ahead).
+#     If path and file exist then the parameters in the file will override the defined
+#     parameters. File parameters format per line:<param_name>: <param value>
+#     Note: Space char after the colon is mandatory
+#   Command line arguments override:
+#     General format per argument is:--<param_name>=<param_value>
+#     those parameters will override the defined parameters and the file parameters.
+#   Override input config file:
+#     User can override the input file by using:--conf_file=<new_file_path>
+#
+# More examples in bbfs/examples/params/rb
+
 require 'optparse'
 require 'yaml'
 
-# A project parameters module.
-# 1. Parameter definition (lowest precedence) - requires phases:
-#    e.g. Params.parameter '<param_name>', <param value>, '<description>'
-#    <param value> must not be nil (since it sets teh type of the parameter).
-#    <param value> supported types are String, Integer(Fixnum), Float, true\false.
-#    <param value> type will define the type of the parameter throughout the entire
-#    execution life time.
-# 2. Params.Init
-#    implementation of the init sequence which allows override of
-#    defined parameters with new values through input file and\or command line args.
-#    Note that only defined parameters can be overridden. If new parameters are passed
-#    through file\command line, then an error will be raised.
-#    2.1 Input config file (medium precedence).
-#        Default configuration input file path is: '~/.bbfs/conf/<executable name>.conf'
-#        This path can be overridden by the command line arguments (see 2.2)
-#        If file path exists then the parameters in the file
-#        will override the defined parameters.
-#        Parameters format per line:<param_name>: <param value>
-#        Note: Space char after the colon is mandatory
-#    2.2 Command line arguments (highest precedence):
-#        General format per argument is:--<param_name>=<param_value>
-#        those params will override the defined and file values
-#        2.2.1 Override input config file
-#              User can override the input file by using:--conf_file=<new filepath>
-#    3.4 Type check:
-#        As mentioned before, types are set at the definition phase (nil is not acceptable).
-#        Therefore, file and command line params should be of the same type as the defined
-#        type, or else an error will be raised.
-#        There is one exception to this rule, and that is, when the defined parameter
-#        is a Float e.g. 3.2 and the user overrides with an integer e.g. 5 (either
-#        through file or command line), than there will be a casting
-#        to 5.0 (original types do not change at any case).
 module BBFS
   module Params
-    @params_initialized = false
 
-    # setter. Mainly for testing.
-    def Params.params_initialized= flag
-      @params_initialized = flag
+    # Represents a parameter.
+    class Param
+      attr_accessor :name
+      attr_accessor :value
+      attr_accessor :desc
+      attr_accessor :type
+
+      # value_type_check method:
+      # 1. Check if member:'type' is one of:Integer, Float, String or Boolean.
+      # 2. input parameter:'value' class type is valid to override this parameter.
+      # 3. Return value. The value to override with correct type. A cast from integer to Float will
+      #    be made for Float parameters which are set with integer values.
+      # 4. Check will be skipped for nil value.
+      def value_type_check value
+        case( @type )
+          when 'Integer' then
+            if not @value.nil?
+              if not ((value.class.eql? Integer) or
+                  (value.class.eql? Fixnum))
+                raise "Parameter:'#{@name}' type:'Integer' but value type to override " \
+                      "is:'#{value.class}'."
+              end
+            end
+          when 'Float' then
+            if not @value.nil?
+              if not value.class.eql? Float
+                if not ((value.class.eql? Integer) or
+                    (value.class.eql? Fixnum))
+                  raise "Parameter:'#{@name}' type:'Float' but value type to override " \
+                        "is:'#{value.class}'."
+                else
+                  return value.to_f
+                end
+              end
+            end
+          when 'String' then
+            if not @value.nil?
+              if not value.class.eql? String
+                raise "Parameter:'#{@name}' type:'String' but value type to override " \
+                      "is:'#{value.class}'."
+              end
+            end
+          when 'Boolean' then
+            if not @value.nil?
+              if not((value.class.eql? TrueClass) or (value.class.eql? FalseClass))
+                raise "Parameter:'#{@name}' type:'Boolean' but value type to override " \
+                      "is:'#{value.class}'."
+              end
+            end
+          else raise "Parameter:'#{@name}' type:'#{@value.class}' but parameter " \
+                       "type to override:'#{value.class}' is not supported. " + \
+                       "Supported types are:Integer, Float, String or Boolean."
+        end
+        return value
+      end
+
+      # supported types are: String, Integer, Float and Boolean
+      def initialize name, value, type, desc
+        @name = name
+        @type = type
+        @desc = desc
+        @value = value_type_check value
+        puts "initialize new param:  name=#{name}  value=#{value}  type=#{type}  desc=#{desc}"
+      end
+    end
+
+    # The globals data structure.
+    @globals_db = Hash.new
+
+    def Params.raise_error_if_param_does_not_exist name
+      if not @globals_db[name]
+        raise "before using parameter:'#{name}', it should first be defined through Param module methods:" \
+               "Params.string, Params.integer, Params.float or Params.boolean."
+      end
+    end
+
+    def Params.raise_error_if_param_exists name
+      if @globals_db[name]
+        raise "Parameter:'#{name}', can only be defined once."
+      end
+    end
+
+    # Read global param value by other modules.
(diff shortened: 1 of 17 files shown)


test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+
+require_relative '../../lib/params.rb'
+
+module BBFS
+  module Params
+    module Spec
+      describe 'Params::parameter' do
+        it 'should raise error when parameter is defined as nil' do
+          Params.params_initialized = false
+          expect { Params::parameter('tmp', nil, 'tmp nil def') }.should raise_error \
```
kolmanv: This is not good. In many cases I want parameter as nil and in specific runs I will set it to some path or string or number.
The nil value let me do:
copy(Params.path) unless Prams.path  # If Params.to is nil, copy will not execute.

kolmanv: The meaning of the value is default value which can be nil.

test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+require 'yaml'
+
+require_relative '../../lib/params.rb'
+
+module BBFS
+  module Params
+    module Spec
+      describe 'Params::parameter' do
+        it 'should raise error when parameter is defined as nil' do
+          Params.params_initialized = false
```
kolmanv: Why this line is needed?

test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+              "loaded yml param:'tmp3' which does not exist in Params module."
+        end
+
+        it 'should raise error when yml parameter type is different then definition' do
+          Params.params_initialized = false
+          # string to other
+          Params::parameter('tmp4str', 'string_value', 'tmp4 def')
+          expect { Params::read_yml_params StringIO.new 'tmp4str: strr' }.should_not raise_error
+          expect { Params::read_yml_params StringIO.new 'tmp4str: 4' }.should raise_error \
+              "loaded yml param:'tmp4str' has different value type:'Fixnum'" + \
```
kolmanv: This is problematic because if I want a a string with value '4' it is a valid string.

kolmanv: Maybe make it a warning but allow this.

test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+          expect { Params::read_yml_params StringIO.new 'tmp4true: 4' }.should raise_error \
+              "loaded yml param:'tmp4true' has different value type:'Fixnum'" + \
+                " then the defined parameter type:'TrueClass'"
+          expect { Params::read_yml_params StringIO.new 'tmp4true: 4.5' }.should raise_error \
+              "loaded yml param:'tmp4true' has different value type:'Float'" + \
+                " then the defined parameter type:'TrueClass'"
+          expect { Params::read_yml_params StringIO.new 'tmp4true: true' }.should_not raise_error
+          expect { Params::read_yml_params StringIO.new 'tmp4true: false' }.should_not raise_error
+
+          # false to other
```
kolmanv: You mean boolean to other?

test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+          Params.parameter 'tmp', 1, 'tmp'
+          Params.tmp.should eq 1
+        end
+
+        it 'should raise an error when trying to define parameter after init phase' do
+          Params.params_initialized = true
+          expect { Params::parameter('tmp2', 'dummy', 'tmp2 not ok') }.should raise_error \
+              "parameters already initialized. No new definitions are allowed."
+        end
+      end
```
kolmanv, line 36: Please add tests:
1) When defining a parameter twice raise exception.

Bad:
Params.parameter('ku', 1, 'ku desk')
Params.parameter('ku', 2, 'ku desk2')

Good:
Params.parameter('ku', 1, 'ku desk')
Parameter.ku = 4  # change the default value.

test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+          # from float to other types.
+          Params::parameter('tmp7', 8.9, 'tmp7 def')
+          # Casting fix num to float
+          expect { Params::parse_command_line_arguments ['--tmp7=9'] }.should_not raise_error
+          expect { Params::parse_command_line_arguments ['--tmp7=3.4'] }.should_not raise_error
+          expect { Params::parse_command_line_arguments ['--tmp7=ff'] }.should raise_error
+          expect { Params::parse_command_line_arguments ['--tmp7=true'] }.should raise_error
+          expect { Params::parse_command_line_arguments ['--tmp7=false'] }.should raise_error
+
+          # from TrueClass to other types.
```
kolmanv, line 164: There is no boolean? :)

test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+          File.stub(:exist?).and_return true
+          File.stub(:open).and_return StringIO.new 'init_param: yml'
+          Params.init ['--conf_file=dummy_file', '--init_param=command-line']
+          Params.init_param.should eq 'command-line'
+        end
+
+        it 'init should override parameters with file only' do
+          Params.params_initialized = false
+          Params::parameter('init_param', 'code', '')
+          File.stub(:exist?).and_return true
```
kolmanv, line 253: Very nice use of stubs.

test/params/params_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+
+        it 'init should not override any parameters' do
+          Params.params_initialized = false
+          Params::parameter('init_param', 'code', '')
+          File.stub(:exist?).and_return false
+          Params.init []
+          Params.init_param.should eq 'code'
+        end
+
+        it 'init should raise error when defining new param after init' do
```
kolmanv: Why is this nescessary?

lib/params.rb
```diff
@@ -1,38 +1,173 @@
 require 'optparse'
 require 'yaml'
-# A simple params module.
+
+# A project parameters module.
+# 1. Parameter definition (lowest precedence) - requires phases:
+#    e.g. Params.parameter '<param_name>', <param value>, '<description>'
+#    <param value> must not be nil (since it sets teh type of the parameter).
```
kolmanv: teh => the
I would not constraint this.
Better make 3 definition functions:
Parameter.int('an_integer_flag', 33, 'int description')
Parameter.string('a_string_flag', '33', 'string description')
Parameter.float('a_float_flag', 33.33, 'float description')

# Those not mandatory to implement now:

Parameter.array('an_array_flag', [33, 321, 1], 'array description')
Parameter.hash('a_hash_flag', 33, 'hash description')

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+#    <param value> supported types are String, Integer(Fixnum), Float, true\false.
+#    <param value> type will define the type of the parameter throughout the entire
+#    execution life time.
+# 2. Params.Init
+#    implementation of the init sequence which allows override of
+#    defined parameters with new values through input file and\or command line args.
+#    Note that only defined parameters can be overridden. If new parameters are passed
+#    through file\command line, then an error will be raised.
+#    2.1 Input config file (medium precedence).
+#        Default configuration input file path is: '~/.bbfs/conf/<executable name>.conf'
```
kolmanv: I would consider setting the default value here to 'nil', usually I will run with command line arguments.
Better ask Alexey, our tester.

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+        /--conf_file=([a-zA-Z0-9\-_\.]+)/ =~ argument
+        if not $1.nil?
+          Params.conf_file = $1
+          break
+        end
+      end
+    end
+
+
+    def Params.print_all_project_parameters
```
kolmanv: print_all_project_parameters => print_all_parameters
This is not all project parameters, but only part because some files are not included/required by all executables.

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+      puts "\nProject initialized parameters:"
+      puts "---------------------------------"
+      names = Params.instance_variables  # Define all available parameters
+      counter=0
+      names.each do |name|
+        tmp_name = name.to_s
+        tmp_name.slice!(0)
```
kolmanv: What does this do?

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+        counter += 1
+        puts "#{counter}: #{tmp_name}=#{Params.instance_variable_get name}"
+      end
+      puts "---------------------------------\n\n"
+    end
+
+    # Initializes the project parameters
+    def Params.init args
+      Params.parameter 'conf_file', \
```
kolmanv: I would start with parse command line arguments first.
Then check if yml file is overridden or not, actually you don't have to, just check Params.conf_file.
The downside here is that you cannot override some parameters now.
You can help yourself by adding auxiliary data to each parameter.
But I don't like the regexp you use to find out whether the config_file was overriden.
Think what is simpler. and easy to maintain.

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+    def Params.executable_name
+      /([a-zA-Z0-9\-_\.]+):\d+/ =~ caller[caller.size-1]
+      return $1
+    end
+
+    #Auxiliary method to check if config file has been provided thorough command line.
+    def Params.override_config_file_path args
+      args.each do |argument|
+        # check parameter format: '--conf_file=<file_path>'
+        /--conf_file=([a-zA-Z0-9\-_\.]+)/ =~ argument
```
kolmanv: You have a parsing parameters function, use it.
Make it return a hash and not write itself directly to the Params module.
Then after config file checks and loads merge the hash to Params module to sustain precedence.

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+      @params_initialized = true
+      Params.print_all_project_parameters
+    end
+
+    # Define new project parameters.
+    # After Params.init is called then this method will raise an error.
+    # Default value can not be nil.
+    def Params.parameter(name, value, description)
+      if value.nil?
+        raise "parameter:'#{name}' value can not be nil."
```
kolmanv: I think we should allow nil values.

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+        #overriding existing parameter
+        if (existing_value.class.eql? TrueClass) or (existing_value.class.eql? FalseClass)
+          # true\false case
+          new_value_upcase = value.to_s.upcase
+          if new_value_upcase.eql? 'TRUE'
+            self.instance_variable_set '@' + name, true
+          elsif new_value_upcase.eql? 'FALSE'
+            self.instance_variable_set '@' + name, false
+          else
+            raise "Defined value type is boolean but overriding value is not boolean."
```
kolmanv: Write what are the options, i.e., 'true' or 'false' only
 or add more 't' 'f' 'y' 'n' 'yes' 'no' ......

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+            #cast from Fixnum to Float
+            Params.parameter param_name, proj_params[param_name] + 0.0, ''
+          else
+            if (((value.class.eql? TrueClass) and (proj_params[param_name].class.eql? FalseClass)) \
+               or \
+               ((value.class.eql? FalseClass) and (proj_params[param_name].class.eql? TrueClass)))
+              #true\false case
+              Params.parameter param_name, proj_params[param_name], ''
+            else
+              #not cast from fixnum to float and not true\false case
```
kolmanv: fixnum and int are almost the same if we use them as parameters.
consider leaving only one.

kolmanv:
Good work. If you 100% sure how to fix everything please merge after you fix.

test/params/params_spec.rb
yarondbb: The reason is to a simple type defiition and check.
The override system should check the types. and nil allows override with any value type.

One solution is to define a string value as '' and ask:  copy(Params.path) unless Prams.path.empty?
Other solution is to have the type added by the user:
Params.parameter my_param, nil, 'desc', 'String'

Also, I do not like parameters defined with no value, I think it might lead to errors.

test/params/params_spec.rb
yarondbb: all tests are added with this line to make the test portable.
Since this flag might be changed by a different test I want to make sure it is in its false mode.
There are other solutions such as make sure that if other test sets it to rtue (like the init tests then it would also set it back to false. To use this approach?

test/params/params_spec.rb
yarondbb: Ok I will make it same as the command parsing which cast all types to string is this is the defined type

test/params/params_spec.rb
yarondbb: there is no boolean in ruby :) only TrueClass or FalseClass

test/params/params_spec.rb
yarondbb, line 36: Ok I will change a bit the desing to do that

test/params/params_spec.rb
yarondbb, line 164: no.. TestClass or FalseClass

test/params/params_spec.rb
yarondbb: Exactly to address your comment before reagring re-definition.
Your advice is better:
1. allow only one call to Param.parameter for a given parameter)
2. let the file and command line override use the set instance param and not use Param.parameter.

I will implement as you suggest

lib/params.rb
yarondbb: Let's talk about the nil contraint

lib/params.rb
yarondbb: If file is not there it would not be used.
I think it is good to set a defualt value to this parameter (help to orginaize the execution in our dirs structure) under .bbgs

lib/params.rb
yarondbb: right. mybe use: print_global_params

lib/params.rb
yarondbb: opps forgot documenting..

lib/params.rb
yarondbb: Ok. I liked your suggestion to use the parser and save the param=value it creates, then check for file override and use it, then use the command line parser data to override again

lib/params.rb
yarondbb: Ok. good approach. will use it

lib/params.rb
yarondbb: Let's talk about it

lib/params.rb
yarondbb: ok

lib/params.rb
yarondbb: I will look into it

yarondbb:
Ok. I will talk with you if not 100% sure of anything (like the nil and fixnum\int and boolean)

test/params/params_spec.rb
kolmanv: This is good.

test/params/params_spec.rb
kolmanv: Still we may use boolean because all know what it means and inside we use TrueClass or FalseClass.

test/params/params_spec.rb
kolmanv: ok

lib/params.rb
kolmanv: Problem will happen when a person have the config file (which is very likely) but does not intend to use it. Then for passive behavior of not using the file he will have to do an active behavior of setting the flag to empty string or nil. This is counter intuitive better to use it if he specified in the flag to use it.
Yes it helps to organize the dirs but I think we should have a empty string/nil value here anyway.

lib/params.rb
kolmanv: ?print_params

lib/params.rb
kolmanv: You will have to change Alexey parse function a bit for that.

lib/params.rb
kolmanv: I mean we never change the parameters once they are defined.

lib/params.rb
```diff
@@ -1,38 +1,173 @@
...
+
+      # Check if parameters already initialized
+      if @params_initialized
+       raise "parameters already initialized. No new definitions are allowed."
+      end
+
+      existing_value = self.instance_variable_get '@' + name
+      if existing_value.nil?
+        # new parameter
+        self.instance_variable_set '@' + name, value
```
kolmanv: I'm thinking of changing this approach, while it is comfortable to use it seems kinda wrong because we have more and more member methods and functions, i.e., names which parameters cannot be like I cannot make a parameter called parameter:
Params.parameter('parameter', 3, 'desc')
Because it will override the definition function (parameter method).
I think of using simple hash and set and get methods, maybe with []
Params[:parameter]  # with symbols
Params['file_path']  # with strings
Don't know what better.

PR #70, opened 2012-07-18
Apply background

genadyp:
Added option to run content_server and backup_server in background.
Secondary: run_in_background will issue output to log and stdout


bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
   require 'content_server'
 end
+include BBFS
+
+Params['log_debug_level'] = 1
+
+#--------------------------
+#  Parameters
+begin
+  Params.raise_error_if_param_exists  'backup_server_background'
```
yarondbb: This is redundent check since it happens automatically when you try to define new param in the next line

bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
 rescue LoadError
   require 'rubygems'
+  require 'yaml'
+  require 'params'
+  require 'run_in_background'
   require 'content_server'
 end
+include BBFS
+
+Params['log_debug_level'] = 1
```
yarondbb: Better to put this line after line 44 (after init has been called, since init could potentionally override this param value)
or this is your intent? to change the default value to 1 and let the user decide if to override it?

bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
+  # parameters already defined <= running daemon now
+end
+
+#--------------------------
+# Initialization
+Params.init ARGV
+
+if File.exists? Params['backup_server_config']
+  File.open(Params['backup_server_config'], 'r') do |file|
+    Params.read_yml_params file
```
yarondbb: Wrong use of Params.read_yml_params_file. You should put all the parameters of the executable in one yml file and let Params.init handle it. If the file location is not the default location:~/.bbfs/conf/#{Params.executable_name}.conf  (as in your case) you should override it like this:
Params['conf_file'] =  Params['backup_server_config']
And then call init (line 44)

bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
+  end
+end
+Params['log_param_number_of_mega_bytes_stored_before_flush'] = 0
+Params['log_param_max_elapsed_time_in_seconds_from_last_flush'] = 0
+
```
yarondbb: These values have become the defulat log values 0 and 0 now . you can remove lines 51-52

bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
+#--------------------------
+# Functions
+def update_argv_with_config_file config_file_path
+  if File.exists? Params['backup_server_config']
+    config = YAML.load_file config_file_path
+    config.each do |key, value|
+      # format taken from the commnad line parsing of Params
+      # values are escaped for the case that they contain spaces
+      value = %Q{"#{value}"} if value =~ / /
```
yarondbb: What does this line do? ;) need description

bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
+    end
+  end
+end
+
+# prepare ARGV so it can be provided as a command line arguments
+def prepare_argv
+  new_argv = Array.new
+  ARGV.each do |arg|
+    arg_arr = arg.split '='
+    if arg_arr.size == 1
```
yarondbb: We do not allow this case normaly only: --<param>=<value> is expected.
Do you have other flags which do not start with -- which you use in the algorithm?

bin/backup_server
yarondbb: if  background is not used then:
   param_str: this is a string
would be used as 'this is a sting'

If background used, then parms_str would become '\"this is a string\"'
Would it affect the usage of this param? if the string will be used for comparisons for excample..?

bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
+      arg = "\"#{arg}\"" if arg =~ / /
+      new_argv << arg
+    elsif arg_arr.size == 2
+      arg_arr[1] = "\"#{arg_arr[1]}\"" if arg_arr[1] =~ / /
```
yarondbb: Arn't you repeting the same action as in line 62?
So file parameter would get the quotes twice?

bin/backup_server
```diff
@@ -5,13 +5,129 @@
...
+      new_argv << arg_arr.join('=')
+    else
+      Log::warning "ARGV argument #{arg} wasn't processed"
+      new_argv << arg
+    end
+  end
+  ARGV.clear
+  ARGV.concat new_argv
+end
+
```
yarondbb: I want to understand better the usage of the ARGV parameters so that all the parameters handling would be made in init.
there is a lot of ARGV parsing going on..

bin/run_in_background/daemon_wrapper
```diff
@@ -47,45 +43,75 @@ class WrapperDaemon < Daemon
         Process.kill 0, @pid  
       rescue Errno::ESRCH
         # if inner application exited then stop the service
+        BBFS::Log.debug1 "inner app no more running"
```
yarondbb: change " to '

bin/backup_server
genadyp: it's not redundant cause when running in Linux - the process moved to background and it still recognized all parameters that already were defined. 
there are two runnings - first is to run myself in background (here parameters defined), second - actual running in background - here in Linux, in contrast with Windows, process that run in background recognized already defined parameters and second attempt to define parameters cause to unwanted exception.
it's the reason that this code run in begin/cause block

bin/backup_server
genadyp: need to remove 
Params['log_debug_level'] = 1
it's an artefact from development

bin/backup_server
genadyp: should be fixed by defining correct default config file location

bin/backup_server
genadyp: 'this is a sting' !=  '\"this is a string\"'
as we agreed in discussion we will use new array based on the ARGV as a parameter for Run_In_Background.start

bin/backup_server
genadyp: line 62: config file treatment
line 77: command line parameters treatment
it will be better to unite two similar treatments

bin/run_in_background/daemon_wrapper
```diff
@@ -24,20 +22,18 @@ rescue LoadError
...
 # On WindowsXP log can be found under:
 # C:/Documents and Settings/NetworkService/.bbfs/daemon_wrapper_<pid>.log
 BBFS::Params['log_file_name'] = File.join(Dir.home, '.bbfs', "#{File.basename(__FILE__)}_#{Process.pid}.log")
 BBFS::Params['log_param_max_elapsed_time_in_seconds_from_last_flush'] = 1
-if DBG
-  BBFS::Params['log_debug_level'] = 1
-  BBFS::Log.init
-end
+# NOTE change it to receive more debug info
+BBFS::Params['log_debug_level'] = 1
```
yarondbb: Need to remove this line (artifect from development)

bin/run_in_background/daemon_wrapper
```diff
@@ -24,20 +22,18 @@ rescue LoadError
...
+BBFS::Log.init
 
 class WrapperDaemon < Daemon
   def service_main
-    BBFS::Log.debug1 "Wrapper starts: #{ARGV.join(' ')}"
+    BBFS::Log.info "Wrapper starts: #{ARGV.join(' ')}"
     @pid = Process.spawn ARGV.join(" ")
```
yarondbb: change to single quotes

lib/run_in_background.rb
```diff
@@ -299,6 +352,79 @@ def RunInBackground.running? name
...
+    # Returns load path as it provided in command line
+    def RunInBackground.get_load_path
+      load_path = Array.new
+      $:.each do |location|
+        load_path << %Q{-I"#{get_abs_std_path(location)}"}
+      end
+      load_path.join ' '
+    end
+
+    # prepare ARGV so it can be provided as a command line arguments
```
yarondbb: add short desc such as:
remove paramtere:'bg_command=' from ARGV to prevent endless run in the backgroung executions which will crush the machine.
add quotes to all parameters (sinc ruby interpreter removes them in the first execution)

lib/run_in_background.rb
```diff
@@ -299,6 +352,79 @@ def RunInBackground.running? name
...
+      case Params['bg_command']
+        when nil
+          yield b
+        when 'start'
+          # to prevent service enter loop cause of background parameter
+          # all options that points to run in background must be disabled
+          # (for more information see documentation for RunInBackground::start!)
+          Params['bg_command'] = nil
+          RunInBackground.prepare_argv
+
```
yarondbb, line 408: Good technique, will work both for windows and for linux.

lib/run_in_background.rb
```diff
@@ -198,17 +230,27 @@ def RunInBackground.start! name, opts = {}
     # <br>For more information see Win32::Daemon help and examples.
     # <br>No need  to wrap such a script.
     def RunInBackground.start_win32service binary_path, binary_args, name, opts_specific = {}
+      Log.debug1 "executable that should be run as service: #{binary_path}"
+      Log.debug1 "arguments: #{binary_args}"
+      Log.debug1 "specific oprions: #{opts_specific}"
```
yarondbb: typo: oprions --> options

lib/run_in_background.rb
```diff
@@ -189,6 +220,7 @@ def RunInBackground.start_windows binary_path, binary_args, name, opts = {}
     # to run script in background, <br>otherwise an unexpexted result can be received
     def RunInBackground.start! name, opts = {}
       start(File.expand_path($0), ARGV, name, opts)
```
yarondbb, line 221: add description that $0 is the executable name

lib/run_in_background.rb
```diff
@@ -37,14 +37,11 @@ module BBFS
   #    method whether daemon already exists and with <tt>running?</tt> method does it running.
   module RunInBackground
     require 'params'
+    require 'log'
```
yarondbb: Why are the requires in this place? should be at top of file

lib/run_in_background.rb
```diff
@@ -37,14 +37,11 @@ module BBFS
...
-    Params.string 'pid_dir', File.expand_path(File.join(Dir.home, '.bbfs', 'pids')),
-      'Absolute path to directory, where pid files will be stored. ' + \
-      'User should have a read/write permissions to this location. ' + \
-      'If absent then it will be created. ' + \
-      'It\'s actual only for Linux/Mac. ' + \
-      'For more information see documentation on daemons module. ' +\
-      'Default location is: $HOME/.bbfs/pids'
+    Params.string('bg_command', nil, 'Server\'s command. Commands are: start, delete and nil for' \
+                  ' not running in background.')
+    Params.string('service_name', Log.executable_name, 'Background service name.')
```
yarondbb: concider use $0 for executable name with proper comment (cryptic ruby magic..)

lib/run_in_background.rb
```diff
@@ -37,14 +37,11 @@ module BBFS
...
 
     # maximal time to wait untill OS will finish a requested operation
```
yarondbb: typo: untill --> until

yarondbb:
for all files apply the new style guid regarding methods call paranthesis -->
for methods with no params --> no paranthesis are required
for methods with 1 or more params -> use paranthesis

lib/run_in_background.rb
```diff
@@ -61,6 +58,14 @@ module RunInBackground
```
yarondbb, line 57: at line 48:  TIMEOUT = 20
concider making this a Params parameter. Any reason it is not?

lib/run_in_background.rb
```diff
@@ -61,6 +58,14 @@ module RunInBackground
...
+      Params.string 'pid_dir', File.expand_path(File.join(Dir.home, '.bbfs', 'pids')),
+        'Absolute path to directory, where pid files will be stored. ' + \
+        'User should have a read/write permissions to this location. ' + \
+        'If absent then it will be created. ' + \
+        'It\'s actual only for Linux/Mac. ' + \
+        'For more information see documentation on daemons module. ' +\
+        'Default location is: $HOME/.bbfs/pids'
+
       # directory used to store pid files
       PID_DIR = File.expand_path(Params['pid_dir'])
```
yarondbb: Params['pid_dir']) already exanded the dir. 
Concider using  Params['pid_dir']) directly instead of using PID_DIR

lib/run_in_background.rb
```diff
@@ -115,9 +120,15 @@ module RunInBackground
```
yarondbb, line 117: at line: 79
    ::FileUtils.mkdir_p PID_DIR
concider remove redundant :: and use:
     FileUtils.mkdir_p PID_DIR

yarondbb, line 117: at line 95: typo: Managment --> Management

yarondbb:
great work Genady Kolman!
After all is fixed you can merge the files

lib/run_in_background.rb
genadyp, line 117: ::FileUtils.mkdir_p means search for method in FileUtils core gem (correct place)
FileUtils.mkdir_p means search for method in BBFS::FileUtils (incorrect place)

bin/backup_server
kolmanv: The prepare argv should work to any command line, not specifically to our Params.

bin/backup_server
kolmanv: Added comments.

bin/run_in_background/daemon_wrapper
kolmanv: Done.

bin/backup_server
kolmanv: After I moved those lines to lib/... this problem no longer produced, because require works only once anyway.
Done.

bin/backup_server
kolmanv: Removed.

bin/backup_server
kolmanv: Removed.

bin/backup_server
kolmanv: Removed. No needed.

bin/backup_server
kolmanv: United.

bin/run_in_background/daemon_wrapper
kolmanv: Done.

bin/run_in_background/daemon_wrapper
kolmanv: Done.

lib/run_in_background.rb
kolmanv: Added comment.

lib/run_in_background.rb
kolmanv, line 408: Thanks.

lib/run_in_background.rb
kolmanv: Done.

lib/run_in_background.rb
kolmanv, line 221: Done.

lib/run_in_background.rb
kolmanv: Done.

lib/run_in_background.rb
kolmanv: Done.

lib/run_in_background.rb
kolmanv, line 57: Done,

Question: When should we use constant, when should we use params?

lib/run_in_background.rb
kolmanv: Done.

lib/run_in_background.rb
kolmanv, line 117: Fixed typo.

PR #79, opened 2012-08-03
Network Queue

kolmanv:


lib/networking/nqueue.rb
```diff
@@ -0,0 +1,59 @@
...
+module BBFS
+  module Networking
+
+    # NQueue is bi-directional queue which transfers object using network (tcp/ip).
+    # The queue has two working modes:
+    # a) As TCP client - when initialized with remote port + host.
+    # b) As TCP server - when initialized with port only.
+    # In both cases after the TCP connection is esteblished the user can push objects to the queue
+    # and pop objects from it which comes from the other side.
+    # The qeueu interface and behaiviour is the same as basic Queue.
```
yarondbb, line 14: Few typos (dictionary) in the comments

lib/networking/tcp.rb
```diff
@@ -99,6 +103,12 @@ def initialize(host, port, obj_clb=nil, reconnected_clb=nil)
         end
       end
 
+      def addr_info
+        Log.info(@tcp_client.addr.inspect) if @tcp_client
+        return @tcp_client.addr if @tcp_client
+        return '#{host}:port'
```
yarondbb, line 109: 1. Where does @tcp_client gets updated?
2. Should it be: "#{host}:port"  - (" and not ') ?

PR #84, opened 2012-09-24
add platform-specific dependencies to the gem

genadyp:
Fix for the platform-specific gem dependencies.
It fix the following problem that appears when installing content_server from gem on Linux:
Fetching: win32-api-1.4.8.gem (100%)
Building native extensions.  This could take a while...
ERROR:  Error installing content_server:
        ERROR: Failed to build gem native extension.

### DIFF
--- a/ext/run_in_background/mkrf_conf.rb
+++ b/ext/run_in_background/mkrf_conf.rb
@@ -1,4 +1,4 @@
-# this extension used to install platform specific gem dependencies
+# This extension used to install platform specific gem dependencies
 # for more information:
 #     http://stackoverflow.com/questions/4596606/rubygems-how-do-i-add-platform-specific-dependency/10249133#10249133
 #     https://github.com/openshift/os-client-tools/blob/master/express/ext/mkrf_conf.rb
--- a/ext/run_in_background/mkrf_conf.rb
+++ b/ext/run_in_background/mkrf_conf.rb
@@ -1,0 +1,34 @@
+# this extension used to install platform specific gem dependencies
+# for more information:
+#     http://stackoverflow.com/questions/4596606/rubygems-how-do-i-add-platform-specific-dependency/10249133#10249133
+#     https://github.com/openshift/os-client-tools/blob/master/express/ext/mkrf_conf.rb
+
+require 'rubygems'
+require 'rubygems/command.rb'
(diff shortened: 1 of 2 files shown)


ext/run_in_background/mkrf_conf.rb
```diff
@@ -0,0 +1,34 @@
+# this extension used to install platform specific gem dependencies
```
kolmanv: this => This

ext/run_in_background/mkrf_conf.rb
```diff
@@ -0,0 +1,34 @@
...
+# for more information:
+#     http://stackoverflow.com/questions/4596606/rubygems-how-do-i-add-platform-specific-dependency/10249133#10249133
+#     https://github.com/openshift/os-client-tools/blob/master/express/ext/mkrf_conf.rb
+
+require 'rubygems'
+require 'rubygems/command.rb'
+require 'rubygems/dependency_installer.rb'
+#begin
```
kolmanv, line 9: Why you commented those 3 lines?

run_in_background.gemspec
```diff
@@ -15,9 +15,10 @@ Gem::Specification.new do |s|
...
-  s.add_dependency('daemons')
-  s.add_dependency('win32-service')
-  s.add_dependency('sys-uname')
-  # Shouldn't be run from comman line
-  # #s.executables << 'run_in_background/daemon_wrapper'  
+#  Add platform dependant gems via extension
+#    Linux dependencies: daemons
+#    Windows dependencies: win32-service, sys-uname
+  s.extensions = ['ext/run_in_background/mkrf_conf.rb']
+  # Shouldn't be run from comman line, thus commented
```
kolmanv, line 22: Why those two lines are commented?

ext/run_in_background/mkrf_conf.rb
genadyp, line 9: These lines were in the example that I  took as a base.
In the current implementation they are irrelevant.
All this code looks like using advanced options of ruby.
Remains as a commented code to ease solving possible errors caused by future changes.

ext/run_in_background/mkrf_conf.rb
genadyp: done

PR #89, opened 2013-01-05
Process monitoring web interface

kolmanv:


lib/process_monitoring.rb
```diff
@@ -0,0 +1,6 @@
+require 'process_monitoring/monitoring'
+require 'process_monitoring/monitoring_info'
+require 'process_monitoring/thread_safe_hash'
+
+module ProcessMonitoring
```
genadyp, line 5: documentation absent - just a few words what is a module, it's goals, etc.

lib/process_monitoring/monitoring.rb
```diff
@@ -1,86 +1,85 @@
...
-            handle_monitoring_state()
-            @process_variables.inc('monitoring_loop_count')
-          end
+    def initialize(process_variables)
+      super(false)
+      @passed_time_dur = 0
+      @process_variables = process_variables
+      @thread = Thread.new do
+        loop do
+          Log.info 'Sleeping in process monitoring.'
```
genadyp: Should it be Log,info or Log.debug ?
Do we have here some kind of performance issue with logs?

lib/process_monitoring/monitoring.rb
```diff
@@ -1,86 +1,85 @@
...
-          :password => Params['monitoring_password'],
-          :body => get_monitoring_state_body(),
-          :subject => 'BBFS Current Monitoring State notification',
-        })
-      end
+    def send_monitoring_state_email
+      SendEmail.send_email({
+                               :to => Params['monitoring_email'],
+                               :from => Params['monitoring_email'],
+                               :password => Params['monitoring_password'],
```
genadyp: May be "to" and "from" mails should be different?
Thus it will be more flexible.
Is there a situation that it different?

lib/process_monitoring/monitoring.rb
```diff
@@ -1,86 +1,85 @@
...
-      def send_monitoring_state_email
-        BBFS::SendEmail.send_email({
-          :to => Params['monitoring_email'],
-          :from => Params['monitoring_email'],
-          :password => Params['monitoring_password'],
-          :body => get_monitoring_state_body(),
-          :subject => 'BBFS Current Monitoring State notification',
-        })
-      end
+    def send_monitoring_state_email
```
genadyp: Do we need this method if we have more general send_log_email(...) ?

lib/process_monitoring/monitoring_info.rb
```diff
@@ -1,28 +1,77 @@
...
+require 'json'
+require 'net/http'
+require 'thin'
+require 'sinatra'
+
+require 'params'
+
+# Set up event machine to exit on ctrl+c.
+EventMachine.schedule do
+  trap("INT") do
```
genadyp, line 11: As far as I remember traping signals available only on Linux machines.
Need to check and if it true then add appropriate comment.

lib/process_monitoring/thread_safe_hash.rb
```diff
@@ -1,33 +1,39 @@
-module BBFS
-  module ThreadSafeHash
-    class ThreadSafeHash
-      def initialize()
-        @hash_data = Hash.new()
-        @mutex = Mutex.new
-      end
+# Simple thread safe hash.
+module ThreadSafeHash
```
genadyp, line 2: Is it a general structure?
May be it shouldn't be under process_monitoring?

lib/process_monitoring.rb
kolmanv, line 5: Done.

lib/process_monitoring/monitoring.rb
kolmanv: Done.
No performance problems, why did you find any?

lib/process_monitoring/monitoring.rb
kolmanv: Don't think so, if you have started the servers, you should get the email, and it is basically from you :)
So for now no need for two different emails.

lib/process_monitoring/monitoring_info.rb
kolmanv, line 11: Will check on windows...

lib/process_monitoring/thread_safe_hash.rb
kolmanv, line 2: For now here, if we need it somewhere else, we will move.

PR #90, opened 2013-02-07
Clean usage

kolmanv:
Now it is very easy to use our program:
ruby -Ilib bin/content_server --conf_file=examples/config_content_server.yml
or
ruby -Ilib bin/backup_server --conf_file=examples/config_backup_server.yml

After this merge I will build new gems and check all works in gems too.


bin/content_server
```diff
@@ -5,25 +5,15 @@
...
-  print "2"
   require 'params'
-  print "3"
   require 'run_in_background'
-  print "4"
   require 'content_server'
-  print "5"
 rescue LoadError
-  print "6"
   require 'rubygems'
```
gshilin: If you know, that 'rubygems' will help to load other gems, why wouldn't you load if first (maybe in begin...end), and then load others? Wouldn't it be simpler code?

examples/config_backup_server.yml
```diff
@@ -0,0 +1,23 @@
+# Content server parameters
```
gshilin, line 1: It is generally bad idea to have configuration with hardcoded values. It is much better to keep an example of this file (even with real data) and to permit everyone to copy it and to change as he wants. Otherwise somebody will one day commit his changes to local setup and you'll propagate it to production. The same is true also for config_content_server.yml

lib/content_server.rb
```diff
@@ -27,8 +25,6 @@ module ContentServer
   Params.integer('backup_file_listening_port', 4444, 'Listening port in backup server for files')
   Params.path('content_data_path', '~/.bbfs/var/content.data',
```
gshilin: I believe you wanted to use here (in this file, I mean) values from configuration files.

lib/params.rb
```diff
@@ -119,6 +121,13 @@ def value_type_check(value)
                       "is:'#{value.class}'.")
             end
           end
+        when 'Complex' then
+          if not @value.nil?
```
gshilin: "if not" in plain English is "unless"

lib/params.rb
```diff
@@ -185,6 +198,12 @@ def Params.override_param(name, value)
       existing_param.value = value.to_s
     elsif existing_param.type.eql?('Path')
       existing_param.value = File.expand_path(value.to_s)
+    elsif existing_param.type.eql?('Complex')
+      if value.class.eql?(Hash) || value.class.eql?(Array)
```
gshilin, line 202: Why are you expecting value to be really Hash? Have you ever heard of "Duck typing"?
There are different kinds of hashes, some of them support order, some enable keys to be string or symbol, but they are still hashes.
The same regarding Arrays. Actually File behaves exactly like Array, so why would you disable it?

lib/params.rb
```diff
@@ -253,18 +279,26 @@ def Params.init(args)
...
 
     # Prints help and parameters if needed.
-    puts @help_messages unless @help_messages.empty?
+    if @show_help_and_exit
+      # Print parameters + description and exit.
+      puts "Full list of parameters:"
+      Params.each do |name, param|
+        puts "--#{name}, #{param.type}, default:#{param.value}\n\t#{param.desc}"
+      end
+      exit
```
gshilin, line 288: That's good idea to return value upon exit, they you'll be able to check exit status from external shell.

lib/params.rb
```diff
@@ -303,7 +337,7 @@ def Params.parse_command_line_arguments(args)
       # Define help command for available options
       # executing --help will printout all pre-defined switch options
       opts.on_tail("-h", "--help", "Show this message") do
-        @help_messages << opts
+        @show_help_and_exit = true
```
gshilin, line 340: In OOP this is usually done via exceptions

bin/content_server
kolmanv: Removed the rescue.

examples/config_backup_server.yml
kolmanv, line 1: This is an example file, it is under the example directory.
It is not being used in production.

lib/params.rb
kolmanv: Done.

lib/params.rb
kolmanv, line 202: Is there a way to check that the class implements Hash or Array interfaces (via duck typing).

lib/params.rb
kolmanv, line 288: The default is correct here (true).

lib/params.rb
kolmanv, line 340: Don't follow, how?

lib/params.rb
gshilin, line 202: value.respond_to?(:[])

lib/params.rb
gshilin, line 340:
``` ruby
begin
  foo
rescue ShowHelpOnlyError
  show_help
  exit
end
```

PR #94, opened 2013-02-25
Round clock testing

kolmanv:


email.gemspec
```diff
@@ -0,0 +1,13 @@
```
gshilin: Why don't you use mailer from Rails?

lib/content_data/content_data.rb
```diff
@@ -483,16 +484,15 @@ def deep_check(instance)
       end
     end
 
-    # @raise [ArgumentError] when instance_check_level is incorrect
     def check_instance(instance)
       case Params['instance_check_level']
```
gshilin, line 493: func = "#{Params['instance_check_level']}_check"
send(func, instance) rescue throw ArgumentError.new "Unsupported check level #{Params['instance_check_level']}"

lib/content_data/content_data.rb
```diff
@@ -553,6 +553,166 @@ def get_query(variable, params)
       result_index
     end
 
-    private :shallow_check, :deep_check, :check_instance, :get_query
+    private :shallow_check, :deep_check, :check_instance
```
gshilin, line 561: Is it your code convention to supply a list of functions?
I've never seen it before. Just "private" and afterwards all private functions' definitions.

lib/content_data/content_data.rb
```diff
@@ -553,6 +553,166 @@ def get_query(variable, params)
...
+  # * deep - can take more time, in addition to shallow recalculates hash sum.
+  # @param [Hash] params hash of parameters of validation, can be used to return additional data.
+  #
+  #     Supported key/value combinations:
+  #     * key is <tt>:failed</tt> value is <tt>ContentData</tt> used to return failed instances
+  # @return [Boolean] true when index is correct, false otherwise
+  # @raise [ArgumentError] when instance_check_level is incorrect
+  def validate(params = nil)
+    # used to answer whether specific param was set
+    param_exists = Proc.new do |param|
```
gshilin, line 578: Add "params" as parameter and define this function outside of #validate.
There is no need to do it again and again...

lib/content_data/content_data.rb
```diff
@@ -553,6 +553,166 @@ def get_query(variable, params)
...
+      !(params.nil? || params[param].nil?)
+    end
+
+    # used to process method parameters centrally
+    process_params = Proc.new do |values|
+      # values is a Hash with keys: :content, :instance and value appropriate to key
+      if param_exists.call :failed
+        unless values[:content].nil?
```
gshilin, line 586: I would do this and next "if" to be one-liners.
params[:failed].add_content values[:content] unless values[:content].nil?
params[:failed].add_instance values[:instance] unless values[:instance].nil?

Short and clear.

lib/content_data/content_data.rb
```diff
@@ -553,6 +553,166 @@ def get_query(variable, params)
...
+        end
+      end
+    end
+
+    is_valid = true
+    instances.each_value do |instance|
+      unless check_instance instance
+        is_valid = false
+
+        unless params.nil? || params.empty?
```
gshilin, line 601: Just add this condition to #process_params, the code here will be cleaner

lib/content_data/content_data.rb
```diff
@@ -553,6 +553,166 @@ def get_query(variable, params)
...
+        Log.warning err_msg
+        false
+      end
+    else
+      false
+    end
+  end
+
+  # @raise [ArgumentError] when instance_check_level is incorrect
+  def check_instance(instance)
```
gshilin, line 652: I've seen this function already on line 487 and ever wrote a comment :)

lib/content_data/content_data.rb
```diff
@@ -553,6 +553,166 @@ def get_query(variable, params)
...
+    if (!exact.empty? && (!from.nil? || !to.nil?))
+      raise ArgumentError 'exact and from/to options are mutually exclusive'
+    end
+
+    result_index = ContentData.new
+    instances.each_value do |instance|
+      is_match = false
+      var_value = instance.instance_variable_get("@#{variable}")
+
+      if exact.include? var_value
```
gshilin, line 706: is_match = (exact.include? var_value) || (from.nil? || var_value > from) && (to.nil? || var_value < to)
and there is no need to initialize it separately tree lines above.

lib/content_data/content_data.rb
```diff
@@ -553,6 +553,166 @@ def get_query(variable, params)
...
+          || (!to.nil? && to.kind_of?(Numeric.new.class)))
+      raise ArgumentError 'from and to options supported only for numeric values'
+    end
+
+    if (!exact.empty? && (!from.nil? || !to.nil?))
+      raise ArgumentError 'exact and from/to options are mutually exclusive'
+    end
+
+    result_index = ContentData.new
+    instances.each_value do |instance|
```
gshilin, line 702: Why don't you use inject? That's exactly what it does: gets an initial value (ContentData.new in this case) and returns it after all calculations.

lib/email/email.rb
```diff
@@ -0,0 +1,87 @@
+require 'net/smtp'
```
gshilin, line 1: Why don't you use https://github.com/mikel/mail ? That's standard client, tested by millions of users, support gmail (see its wiki page).
There is no need to re-invent a wheel :-1:

lib/file_utils.rb
```diff
@@ -1,3 +1,4 @@
 require 'file_utils/file_utils'
+
 module FileUtils
```
gshilin, line 3: Why do you need this definition here? Don't you have it in the required file?

lib/email.rb
```diff
@@ -0,0 +1,4 @@
+require 'email/email'
```
gshilin, line 1: Why do you need this definition here? Don't you have it in the required file?

lib/file_utils/file_generator/file_generator.rb
```diff
@@ -6,22 +6,25 @@
 require 'params'
 require 'yaml'
 
+require 'log'
+
 module FileGenerator
-  Params.string('target_path', '~/.bbfs/test_files', 'Represents the target path for files generation')
```
gshilin, line 10: Once I've asked you to check http://ruby-doc.org/stdlib-1.9.3/libdoc/ostruct/rdoc/OpenStruct.html

email.gemspec
kolmanv: Why not use the simple Mail gem? ...As Action Mailer now uses the Mail gem...
There is an option to use Log4r as mailing feature (maybe wrong approach, but seems for testing server is ok).

PR #102, opened 2013-03-29
Testing prepare merge to master

kolmanv:
Yaron, Gena, this branch is merge between round_clock_testing and current master.
Please check the merge is ok (only file problematic was content_data.rb).
Other were merged automatically.
After you test/fix, please merge it to master.


lib/content_data/content_data.rb
```diff
@@ -262,6 +263,11 @@ def from_file(filename)
       number_of_instances = lines[i].to_i
       i += 1
       number_of_instances.times {
+        if lines[i].nil?
+          Log.info "lines[i] if nil !!!, Backing filename: #{filename} to #{filename}.bad"
+          FileUtils.cp(filename, "#{filename}.bad")
+          Log.info lines[i].join("\n")
+        end
```
yarondbb, line 270: When content dat ais empty then number of contents and instances is 0 and 0. Is this concidered bad?
I think we need to set an empty data base and return

lib/content_data/content_data.rb
```diff
@@ -420,6 +425,149 @@ def deep_check(instance_info)
...
+        Log.warning err_msg
+      end
+    else
+      is_valid = false
+      err_msg = "Indexed file #{path} doesn't exist"
+      Log.warning err_msg
+    end
+    is_valid
+  end
+
```
yarondbb: Gena, this is your part, pls see if this merge is correct

lib/content_data/content_data.rb
kolmanv, line 270: This implementation considers that there are at least two lines in the code, which may be 0 and 0.
If the file has only one 0 or none at all then file is not well formatted.
Lets discuss this verbally

lib/content_data/content_data.rb
genadyp: After the merge there is duplication of code: see lines 324 and 512

PR #169, opened 2013-06-15
Issue149

AlexeyNemytov:
Hi Guys 
Please review the changes made 
1. file_monitoring.rb - 
     a. Add initialize - to read files from content dir during init phase ln 15 - 29
     b. ln 58 - modify DirStat initialization
2. monitor_path.rb
     a. Modify FileStat and DirStat initialize to get state and content cache
     b.ln 36 and ln 148
     c. Check if files exist and mark them as stable  - ln 294
3.  dynamic_content_data.rb 
     a. Modify initiliaze ln 13  - 19  to add existing files to last_content_data

### DIFF
--- a/lib/content_data/dynamic_content_data.rb
+++ b/lib/content_data/dynamic_content_data.rb
@@ -1,4 +1,5 @@
 require 'thread'
+require 'params'
 
 module ContentData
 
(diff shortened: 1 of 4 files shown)


bbfsdev:
@AlexeyNemytov Great work!!!
Talked on phone.

AlexeyNemytov:
@kolmanv  
Changes made according to the comments 
Great week everybody .

yarondbb:
great work Alexey!
Guys, I have few questions regarding the algorithm:
1. When file is found in the cache(path match only), do we also need to check the size and modification time to be sure the file data is up to date?
2. The cache will always be queried for new files? even after initialization phase?
3. Let's concider to make the path cache concept as part of ContentData. Then you will only need to query ContentData if path exists.

PR #180, opened 2013-06-22
issue 176-Process monitoring - For all queues, remove thread of queue.si...

yarondbb:


lib/content_server/file_streamer.rb
```diff
@@ -48,49 +48,48 @@ def initialize(send_chunk_clb, abort_streaming_clb=nil)
...
-
     def copy_another_chuck(checksum)
       @stream_queue << [:COPY_CHUNK, checksum]
+      if Params['enable_monitoring']
+        Params['process_vars'].set('File Streamer queue', @stream_queue.size)
+      end
     end
 
     def start_streaming(checksum, path)
       @stream_queue << [:NEW_STREAM, [checksum, path]]
```
yarondbb, line 66: Sholud we move to 'push' as a convention?

yarondbb:
The second commit is issue 174 development. I have added it to issue 176 since I was blocked by it.

bin/backup_server
```diff
@@ -25,12 +25,23 @@ rescue SystemExit, SignalException => exc
   # TODO (genadyp) do we need to trap signals by types?
   puts("Interrupt or Exit happened in #{Params['service_name']} server: #{exc.class}, " +
                 "stopping process.\nBacktrace:\n#{exc.backtrace.join("\n")}")
+  puts("Force writing local content data to #{Params['local_content_data_path']}.")
+  ContentServer::local_dynamic_content_data.last_content_data.to_file(ContentServer::tmp_content_data_file)
+  sleep(0.1)  # Added to prevent mv access issue
```
bbfsdev, line 30: What is mv access issue?
We just have to make sure the file is closed.
I think sleep is not needed here.

bin/backup_server
```diff
@@ -25,12 +25,23 @@ rescue SystemExit, SignalException => exc
   # TODO (genadyp) do we need to trap signals by types?
   puts("Interrupt or Exit happened in #{Params['service_name']} server: #{exc.class}, " +
                 "stopping process.\nBacktrace:\n#{exc.backtrace.join("\n")}")
+  puts("Force writing local content data to #{Params['local_content_data_path']}.")
+  ContentServer::local_dynamic_content_data.last_content_data.to_file(ContentServer::tmp_content_data_file)
```
bbfsdev, line 29: Dynamic content data is not really thread safe... I have bugs.

bin/backup_server
```diff
@@ -25,12 +25,23 @@ rescue SystemExit, SignalException => exc
...
+  ::FileUtils.mv(ContentServer::tmp_content_data_file, Params['local_content_data_path'])
+
+
   Log.error("Interrupt or Exit happened in #{Params['service_name']} server: #{exc.class}, " +
             "stopping process.\nBacktrace:\n#{exc.backtrace.join("\n")}")
   Log.flush
   exit
 rescue Exception => exc
   puts("Exception happened in #{Params['service_name']} server: #{exc.class}:#{exc.message}\nBacktrace:\n#{exc.backtrace.join("\n")}")
+  puts("Force writing local content data to #{Params['local_content_data_path']}.")
```
bbfsdev, line 40: Put those 3 lines in a function. You are using them 3 times.

lib/content_server/content_server.rb
```diff
@@ -106,46 +113,43 @@ def run_content_server
...
-            Log.info("process_vars:Live File obj cnt:#{file_count}")
-            Log.info("process_vars:Live Dir obj cnt:#{dir_count}")
-            Log.info("process_vars:Live Content data obj cnt:#{content_count}")
-            Params['process_vars'].set('Live objs cnt', total_obj_count)
-            Params['process_vars'].set('Live String obj cnt', string_count)
-            Params['process_vars'].set('Live File obj cnt', file_count)
-            Params['process_vars'].set('Live Dir obj cnt', dir_count)
-            Params['process_vars'].set('Live Content data obj cnt', content_count)
-            last_data_flush_time = Time.now
+          sleep(Params['process_vars_delay'])
```
bbfsdev, line 116: What is this sleep for?
All variables should be updated only when they change, not in one global place.
So this sleep is actually not needed, maybe for Time.now only.

lib/content_server/content_server.rb
```diff
@@ -106,46 +113,43 @@ def run_content_server
...
+          ::ContentServer::Globals.process_vars.set('time', Time.now)
+          Log.debug3("process_vars:monitoring queue size:#{monitoring_events.size}")
+          ::ContentServer::Globals.process_vars.set('monitoring queue', monitoring_events.size)
```
bbfsdev, line 119: Those queues should be updated where they change values.

lib/content_server/globals.rb
```diff
@@ -0,0 +1,10 @@
+require 'process_monitoring/thread_safe_hash'
```
bbfsdev, line 1: Why don't you just use
$process_vars = ThreadSafeHash::ThreadSafeHash.new

$ in Ruby means global variable, this will be shorter.

lib/content_server/queue_copy.rb
```diff
@@ -129,6 +129,10 @@ def initialize(host, port, dynamic_content_data)
                                         method(:reset_copy))
       @local_thread = Thread.new do
         loop do
+          pop_queue = @local_queue.pop
+          if Params['enable_monitoring']
```
bbfsdev, line 133: Consider moving

```
if Params ['enable_monitoring']
  ...
end
```

Inside a function, because you use it in so many places....

This code can look like this:

```
pop_queue = @local_queue.pop
process_vars.update_queue('File Copy Client queue', @local_queue)
```

While in global.rb you will do:

```
$process_vars = ThreadSafeHash::ThreadSafeHash.new
$process_vars.define_singleton_method(:set) { |name, value|
  if Params['enable_monitoring']
    $process_vars.set(name, value)
  end
}
$process_vars.define_singleton_method(:inc) { |name|
  if Params['enable_monitoring']
    $process_vars.inc(name)
  end
}
```

Or even better, just inherit `ThreadSafeHash` and override the inc and set methods such that they will include the `if Params...`

lib/content_server/backup_server.rb
```diff
@@ -36,12 +44,11 @@ def run_backup_server
     # create general tmp dir
     FileUtils.mkdir_p(Params['tmp_path']) unless File.directory?(Params['tmp_path'])
     # init tmp content data file
-    tmp_content_data_file = File.join(Params['tmp_path'], 'backup.data')
+    @@tmp_content_data_file = File.join(Params['tmp_path'], 'backup.data')
```
bbfsdev, line 47: Why are you using here class variables?
I would prefer initializing them in the constructor.
We should consider having  ContentServer and/or BackupServer classes.
We should have a general BBFSServer class which will have all the common stuff including local monitoring and indexing...

lib/content_server/backup_server.rb
```diff
@@ -134,44 +141,40 @@ def run_backup_server
...
-            Log.info("process_vars:Live objs cnt:#{total_obj_count}")
-            Params['process_vars'].set('Live objs cnt', total_obj_count)
-            Log.info("process_vars:Live String obj cnt:#{string_count}")
-            Log.info("process_vars:Live File obj cnt:#{file_count}")
-            Log.info("process_vars:Live Dir obj cnt:#{dir_count}")
-            Log.info("process_vars:Live Content data obj cnt:#{content_count}")
-            Params['process_vars'].set('Live String obj cnt', string_count)
-            last_data_flush_time = Time.now
+          sleep(Params['process_vars_delay'])
+          ::ContentServer::Globals.process_vars.set('time', Time.now)
```
bbfsdev, line 147: All those parameters should be encapsulated in a class.

bin/backup_server
yarondbb, line 30: File should be closed by this line
I got acception saying that the file can not be accessed (it happened few times..)

bin/backup_server
yarondbb, line 40: I suggest move to one exception handling and print its type. simple code the better.

lib/content_server/backup_server.rb
yarondbb, line 47: But now we do not have classes. What to do then?

lib/content_server/backup_server.rb
yarondbb, line 147: We have ::ContentServer::Globals.process_vars
I do not understand

lib/content_server/content_server.rb
yarondbb, line 116: Some of the parameters are changing (queues, FileStat etc..) some are general objects such as: string, total object count etc..
FileStat for example can be put when changed
For the general we would still need this block right?

lib/content_server/content_server.rb
yarondbb, line 119: ok

bin/backup_server
bbfsdev, line 30: Then we should do some other kind of sleep, such that when file is ready we will wake up.

bin/backup_server
bbfsdev, line 40: Sounds good. Share this code between content_server and backup_server.

lib/content_server/backup_server.rb
bbfsdev, line 47: write function and share it

```
def backup_run
...
end
```

```
def content_run
...
end
```

```
def server_exception(server_run_function)
try:
  server_run_function()
catch:
  kuku
end
```

```
server_run_function(backup_run)
or
server_run_function(content_run)
```

lib/content_server/backup_server.rb
bbfsdev, line 147: I mean the code below (20 lines) should be in a separate function.

lib/content_server/content_server.rb
bbfsdev, line 116: Yes, but please put those lines in a function and share it between content and backup.
create a common file such that backup and content server will require it.
this file will have functions that will be shared.

lib/content_server/file_streamer.rb
bbfsdev, line 66: Don't understand the question? What is 'push'?
The usage of operator << ? instead of calling member method?

lib/content_server/globals.rb
```diff
@@ -0,0 +1,10 @@
...
+
+module ContentServer
+  class Globals
+    @@process_vars = ThreadSafeHash::ThreadSafeHash.new
+    def self.process_vars
+      @@process_vars
+    end
+  end
+end
```
bbfsdev, line 10: add newline here.

lib/content_server/backup_server.rb
yarondbb, line 147: why?

lib/content_server/file_streamer.rb
yarondbb, line 66: I mean use: 
 @stream_queue,push([:NEW_STREAM, [checksum, path]])

lib/content_server/globals.rb
yarondbb, line 10: between 9 and 10?

lib/content_server/backup_server.rb
yarondbb, line 47: maybe wait and later (after release) implement a final solution which will include use of class?

bin/backup_server
yarondbb, line 30: wait for File.closed?  
  the program night hang..??

PR #188, opened 2013-06-22
Put correct log for each condition

yarondbb:


bbfsdev:
Looks good.

bbfsdev:
Looks good.

PR #193, opened 2013-07-05
Copy files

yarondbb:


bbfsdev:
I see you handled all the things we talked about + moved the tcp call outside.
Good work, please merge the changes.

yarondbb:
I have enhanced the design so it will use a queue for copy on demend (as you wanted).
This is probalby the fastest way to handle it.
Pls review the new code.
Also I have changed the way backup is checking if file already exists - instead of checking the dynamic content data it is now chaking the actuall destination of the file (so we are not dependent on cycles of monitoring which might be slow)

yarondbb:
@kolmanv  pls review again

PR #195, opened 2013-07-06
issues 187 adn 186

yarondbb:


lib/content_server/backup_server.rb
```diff
@@ -56,8 +56,15 @@ def run_backup_server
 
     # Read here for initial content data that exist from previous system run
     initial_content_data = ContentData::ContentData.new
+    initial_content_data = ContentData::ContentData.new
     content_data_path = Params['local_content_data_path']
-    initial_content_data.from_file(content_data_path) if File.exists?(content_data_path)
+    if File.exists?(content_data_path) and !File.directory?(content_data_path)
+      Log.info("reading initial content data that exist from previous system run from file:#{content_data_path}")
+      initial_content_data.from_file(content_data_path)
+    else
```
genadyp, line 63: Is Params['local_content_data_path']  mandatory?
What happens if it optional and default value does not point to any file?
Do we exit then?
What happens when system run first time and no content_data object exists?

lib/content_server/backup_server.rb
```diff
@@ -56,8 +56,15 @@ def run_backup_server
 
     # Read here for initial content data that exist from previous system run
     initial_content_data = ContentData::ContentData.new
+    initial_content_data = ContentData::ContentData.new
     content_data_path = Params['local_content_data_path']
```
genadyp, line 59: Do we need an expand_path here ?

lib/params.rb
```diff
@@ -288,14 +297,14 @@ def Params.init(args)
     end
 
     # Add parameters to log init messages (used by Log.init if param:print_params_to_stdout is true)
-    @init_debug_messages << 'Initialized executable parameters:'
-    @init_debug_messages << '---------------------------------'
+    @init_info_messages << 'Initialized executable parameters:'
+    @init_info_messages << '---------------------------------'
```
genadyp, line 301: mix of data and view. 
i.e. mix of real info messages with "decoration" messages.
consider, please, whether it good enough.

lib/params.rb
```diff
@@ -288,14 +297,14 @@ def Params.init(args)
...
     counter=0
     @params_data_base.values.each do |param|
       counter += 1
-      @init_debug_messages << "Param ##{counter}: #{param.name}=#{param.value}"
+      @init_info_messages << "Param ##{counter}: #{param.name}=#{param.value}"
     end
-    @init_debug_messages << '---------------------------------'
+    @init_info_messages << '---------------------------------'
```
genadyp, line 307: same as for line 301

lib/content_server/content_server.rb
```diff
@@ -49,7 +49,14 @@ def run_content_server
     # Read here for initial content data that exist from previous system run
     initial_content_data = ContentData::ContentData.new
     content_data_path = Params['local_content_data_path']
-    initial_content_data.from_file(content_data_path) if File.exists?(content_data_path)
+    if File.exists?(content_data_path) and !File.directory?(content_data_path)
+      Log.info("reading initial content data that exist from previous system run from file:#{content_data_path}")
+      initial_content_data.from_file(content_data_path)
+    else
```
genadyp, line 55: same as for backup_server.

genadyp:
@yarondbb 
looks good
view comments, please.

lib/content_server/backup_server.rb
yarondbb, line 63: Yes it is mandatory. If not provided we used to get an exception inside initial_content_data.from_file(content_data_path).
This is exactly why I added a raise (on line 65)
raise("Param:'local_content_data_path':'#{content_data_path}' is a file name which does not exist" +
                " or a directory name")
So that user will get a beeter error description.

lib/content_server/backup_server.rb
yarondbb, line 59: it is done since the type of this param is :'path'

lib/params.rb
yarondbb, line 301: Log look ok I think

PR #196, opened 2013-07-07
issue 182

yarondbb:


bin/backup_server
```diff
@@ -7,11 +7,12 @@
 
 # NOTE this file mainly is a copy of content_server
 
-require 'content_server/server'
+require 'content_server'  #todo: yarondbb need it?
```
genadyp, line 10: As far as i understand this requirement brings all params and required modules specified in lib/content_server.rb
If you remove then, it seems, no need for this require

lib/content_server.rb
```diff
@@ -1,17 +1,2 @@
-require 'content_server/content_server'
```
genadyp, line 1: Before this commit there was enough to get content_server and backup_server by single 
<code>require 'content_server'</code>
now we need to require a specific server:
<code>require 'content_server/content_server'</code>
or
<code>require 'content_server/backup_server'</code>
Need to check that all requre statements to content_server were fixed

lib/content_server/server.rb
```diff
@@ -4,6 +4,21 @@
 
 module ContentServer
```
genadyp, line 5: lib/content_server.rb
looks like a good place for this common code
it's just an advice.
consider, please.

genadyp:
@yarondbb 
looks good
view comments, please.

lib/content_server/server.rb
yarondbb, line 5: lib/content_server.rb has the naming of content server
maybe: lib/server instead of lib/content_server/server?

lib/content_server.rb
yarondbb, line 1: Testing server needed a fix
good catch!

PR #197, opened 2013-07-08
181 bug remove dir

yarondbb:


lib/content_data/content_data.rb
```diff
@@ -171,7 +171,8 @@ def stats_by_location(location)
 
     # removes an instance.
     # removes also the content, if content becomes empty
-    def remove_instance(location)
+    def remove_instance(server, path)
```
genadyp, line 175: could you add a method description including parameters types.
it's difficult to understand what are the method parameters and how to use this method.
see: [YARD Tutorials](http://yardoc.org/guides/index.html)

lib/content_data/content_data.rb
```diff
@@ -181,7 +182,7 @@ def remove_instance(location)
       @instances_info.delete(location)
     end
 
-    def remove_directory(dir_to_remove, server)
+    def remove_directory(server, dir_to_remove)
```
genadyp, line 190: could you add a method description including parameters types.
it's difficult to understand what are the method parameters and how to use this method.
see: YARD Tutorials

genadyp, line 190: is dir_to_remove parameter contains absolute path?
otherwise looks like [L189](https://github.com/bbfsdev/bbfs/blob/5f946a2cfc6375a9e19e446cb82e731635f9ec76/lib/content_data/content_data.rb#L189) can be incorrect in some scenarios
also when  dir_to_remove parameter contains absolute path we can use [starts_with?](http://ruby-doc.org/core-2.0/String.html#method-i-start_with-3F) instead of scan. It  [looks faster](http://mervine.net/ruby-methods-vs-regex)

genadyp:
looks goods.
see comments, please.

Issue #198, opened 2013-07-10
If data has not changed - do not write it to system


yarondbb:
In order to make application as 'light' as posible (memory\CPU\performence),  not to write data to server if this data has not changed.
for example the content data we receive from content server is written to file system even if not changed. Also, local content data is written to content.data\backup.data even if dynamic content data has not changed.

yarondbb:
Alexey finished!

PR #205, opened 2013-07-27
https://github.com/bbfsdev/bbfs/issues/198

AlexeyNemytov:
do not write data to remoteif not changed

### DIFF
--- a/lib/content_server/remote_content.rb
+++ b/lib/content_server/remote_content.rb
@@ -17,6 +17,7 @@ module ContentServer
       @remote_tcp = Networking::TCPClient.new(host, port, method(:receive_content))
       @last_fetch_timestamp = nil
       @last_save_timestamp = nil
+      @last_message = nil
       @content_server_content_data_path = File.join(local_backup_folder, 'remote',
                                                     host + '_' + port.to_s)
       Log.debug3("Initialized RemoteContentClient: host:#{host}   port:#{port}  local_backup_folder:#{local_backup_folder}")
@@ -37,10 +38,15 @@ module ContentServer
       	@last_save_timestamp = Time.now.to_i
         write_to = File.join(@content_server_content_data_path,
                              @last_save_timestamp.to_s + '.cd')
-        FileUtils.makedirs(@content_server_content_data_path) unless \
+        same_message = message.to_s.eql?(@last_message.to_s)  #check if same message received
+        if(!same_message)
+          FileUtils.makedirs(@content_server_content_data_path) unless \
               File.directory?(@content_server_content_data_path)
-        count = File.open(write_to, 'wb') { |f| f.write(message.to_s) }
-        Log.debug1("Written content data to file:#{write_to}.")
+          count = File.open(write_to, 'wb') { |f| f.write(message.to_s) }
+          Log.debug1("Written content data to file:#{write_to}.")
+          @last_message =   message
+        end
+        Log.debug1("No need to write remote content data, it has not changed.")
       else
         Log.debug1("No need to write remote content data, it has not changed.")
       end
--- a/lib/content_server/remote_content.rb
+++ b/lib/content_server/remote_content.rb
@@ -59,9 +59,8 @@ module ContentServer
               File.directory?(@content_server_content_data_path)
           File.open(write_to, 'wb') { |f| f.write(message.to_s) }
           Log.debug1("Written content data to file:#{write_to}.")
-
           @last_content_data_id = message.unique_id   # save last content data ID
-
+        end
         Log.debug1("No need to write remote content data, it has not changed.")
       else
         Log.debug1("No need to write remote content data, it has not changed.")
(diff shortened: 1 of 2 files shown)


lib/content_server/remote_content.rb
```diff
@@ -37,10 +38,15 @@ def receive_content(message)
...
         write_to = File.join(@content_server_content_data_path,
                              @last_save_timestamp.to_s + '.cd')
-        FileUtils.makedirs(@content_server_content_data_path) unless \
+        same_message = message.to_s.eql?(@last_message.to_s)  #check if same message received
+        if(!same_message)
+          FileUtils.makedirs(@content_server_content_data_path) unless \
               File.directory?(@content_server_content_data_path)
-        count = File.open(write_to, 'wb') { |f| f.write(message.to_s) }
-        Log.debug1("Written content data to file:#{write_to}.")
+          count = File.open(write_to, 'wb') { |f| f.write(message.to_s) }
```
genadyp: Do we need count variable here? it looks unused.

lib/content_server/remote_content.rb
```diff
@@ -37,10 +38,15 @@ def receive_content(message)
       	@last_save_timestamp = Time.now.to_i
         write_to = File.join(@content_server_content_data_path,
                              @last_save_timestamp.to_s + '.cd')
-        FileUtils.makedirs(@content_server_content_data_path) unless \
+        same_message = message.to_s.eql?(@last_message.to_s)  #check if same message received
```
genadyp: Is last_message is a ContentData object? If so, suggest to use [ContentData::==](https://github.com/bbfsdev/bbfs/blob/master/lib/content_data/content_data.rb#L204).

same_message is a confusing name, cause there are variables that contains  "message" string and that are ContentData object.
Consider to rename it, for example, to is_same_message, cause this name intends to boolean.
or move check to the "if", one line beneath.

genadyp:
@AlexeyNemytov 
Alexey, add, please short methods description for the methods that you used.
Suggest [YARD](http://rubydoc.info/docs/yard/file/docs/GettingStarted.md) for this purpose.

It's enough to do something like:

``` ruby
# class description...
class Request
# code...
end

class MyWebServer
  # Handles a request
  # @param request [Request] the request object
  # @return [String] the resulting webpage
  def get(request) "hello" end

  # (see #get)
  def post(request) "hello" end
end
```

AlexeyNemytov:
nice one

lib/content_server/remote_content.rb
```diff
@@ -17,6 +17,7 @@ def initialize(dynamic_content_data, host, port, local_backup_folder)
       @remote_tcp = Networking::TCPClient.new(host, port, method(:receive_content))
       @last_fetch_timestamp = nil
       @last_save_timestamp = nil
+      @last_message = nil
```
yarondbb: concider rename it to last_content_data for better readility

yarondbb: saving the entire last content data object is memory consuming (might be few mega bytes in memory for big objects).
I recommend using other "smart" solution to check if last content data has changed.

lib/content_server/remote_content.rb
AlexeyNemytov: changed usage to ContentData::==.

lib/content_server/remote_content.rb
AlexeyNemytov: removed

lib/content_server/remote_content.rb
AlexeyNemytov: @yarondbb 
already commited a new changes , just saw your comments
please check a new code
will think about 'smart' solution :)

yarondbb: Where is th enew code? it should be part of this pull request. Did you push it in same branch?

lib/content_server/remote_content.rb
```diff
@@ -37,15 +53,22 @@ def receive_content(message)
       	@last_save_timestamp = Time.now.to_i
         write_to = File.join(@content_server_content_data_path,
                              @last_save_timestamp.to_s + '.cd')
-        FileUtils.makedirs(@content_server_content_data_path) unless \
+
+        is_same_content = (message==@last_message)  #check if same ContentData received
+        if(!is_same_content)
```
yarondbb: could combine lines 57 and 58 to be:
unless message == @last_message
..

lib/content_server/remote_content.rb
```diff
@@ -37,15 +53,22 @@ def receive_content(message)
...
+          FileUtils.makedirs(@content_server_content_data_path) unless \
               File.directory?(@content_server_content_data_path)
-        count = File.open(write_to, 'wb') { |f| f.write(message.to_s) }
-        Log.debug1("Written content data to file:#{write_to}.")
+          File.open(write_to, 'wb') { |f| f.write(message.to_s) }
```
yarondbb, line 60: I suggest using to_file method of ContentData (message is actually an instance ContentData)

AlexeyNemytov:
@yarondbb 
@genadyp 
Please check new commits

yarondbb:
Greate documentation Alexey!!!
I have merged this code manually (due to many conflicts)

PR #219, opened 2013-08-15
Issue 218 load data

yarondbb:


lib/file_monitoring/file_monitoring.rb
```diff
@@ -41,16 +28,60 @@ def set_event_queue(queue)
...
+        dir_stat.set_event_queue(@event_queue) if @event_queue
+        dir_stat_array.push(dir_stat)
+      }
+
+      #Look over loaded content data if not empty
+      # If file is under monitoring path - Add to DirStat tree as stable with path,size,mod_time read from file
+      # If file is NOT under monitoring path - skip (not a valid usage)
+      unless $local_content_data.empty?
+        Log.info("Start build data base from loaded file")
+        $local_content_data.each_instance() {
```
genadyp: remark: $local_content_data.each_instance {

lib/file_monitoring/monitor_path.rb
```diff
@@ -59,25 +59,26 @@ def self.set_log (log)
     # Checks whether file was changed from the last iteration.
     # For files, size and modification time are checked.
     def monitor
+      #Log.info("Start file monitor")
       file_stats = File.lstat(@path) rescue nil
       new_state = nil
       if file_stats == nil
```
genadyp: file_stats.nil? looks better

lib/file_monitoring/monitor_path.rb
```diff
@@ -59,25 +59,26 @@ def self.set_log (log)
...
         new_state = FileStatEnum::NON_EXISTING
         @size = nil
-        @creation_time = nil
+        #@creation_time = nil
         @modification_time = nil
         @cycles = 0
       elsif @size == nil
```
genadyp: @size.nil? looks better

lib/file_monitoring/monitor_path.rb
```diff
@@ -88,14 +89,13 @@ def monitor
...
+      set_state(new_state)
     end
 
     #  Checks that stored file attributes are the same as file attributes taken from file system.
     def changed?(file_stats)
-      not (file_stats.size == @size &&
-          file_stats.ctime.utc == @creation_time.utc &&
-          file_stats.mtime.utc == @modification_time.utc)
+      !((file_stats.size == @size) &&
+        (file_stats.mtime.to_i == @modification_time))
```
genadyp, line 100: while declaring @modification_time need to comment what is a type of this variable.
otherwise it not so clear that this check is correct and need to go and find variable assignments

lib/file_monitoring/monitor_path.rb
```diff
@@ -137,12 +137,48 @@ class DirStat < FileStat
     #
     # * <tt>path</tt> - File location
     # * <tt>stable_state</tt> - Number of iterations to move unchanged directory to stable state
-    def initialize(path, stable_state = DEFAULT_STABLE_STATE, content_data_cache, state)
+    def initialize(path, stable_state = DEFAULT_STABLE_STATE)
       super
       @dirs = nil
       @files = nil
```
genadyp: absent comment describing types of dirs and files, then it rather difficult to understand how to use them.

lib/file_monitoring/monitor_path.rb
```diff
@@ -137,12 +137,48 @@ class DirStat < FileStat
...
+        file_stat.set_event_queue(@event_queue)
+        file_stat.size = size
+        file_stat.modification_time = modification_time
+        file_stat.state = FileStatEnum::STABLE
+        add_file(file_stat)
+      else
+        # Add Dir to tree if not present. index points to new dir path.
+        dir_stat = @dirs[sub_paths[sub_paths_index]]
+        #create new dir if not exist
+        dir_stat = add_dir(DirStat.new(sub_paths[sub_paths_index], @stable_state)) unless dir_stat
```
genadyp: why do we need to continue here if dir_stat already exists?

lib/file_monitoring/monitor_path.rb
```diff
@@ -137,12 +137,48 @@ class DirStat < FileStat
...
+        dir_stat.state = FileStatEnum::STABLE
+        dir_stat.set_event_queue(@event_queue)
+        # continue recursive call on tree with next sub path index
+        dir_stat.load_instance(sub_paths, sub_paths_index+1, size, modification_time)
```
genadyp, line 185: can we suppose that if parent dir exists then also sub-directories already were added to monitoring?

genadyp:
@yarondbb 
very good and  well commented code.
thank you.

lib/file_monitoring/monitor_path.rb
yarondbb, line 185: I agree that @state and @event_queue are already defined in dir_stat if exists. I will remove the settings.
But the recursive call should continue to propogate in the tree to create the new dirs(optinal) and file.
Concider these 2 paths:
/dir1/dir2/file - this will reate dir 1, dir 2 and file
/dir1/dir2/dir3/file - this will create only dir 3 and file. At dir1 and dir2, it will only do the recursive call to travel the tree

genadyp, line 185: Let consider following scenario:
monitoring path (taken from conf):
/dir1/dir2/
instances (taken from $local_content_data):
/dir1/dir2/dir3/file1
/dir1/dir2/dir3/dir4/dir4/file2

Algorithm (am i understand correctly?)
- [adds /dir1/dir2/ to array of monitoring locations](https://github.com/bbfsdev/bbfs/blob/8f853bb8c481d04e8e57e76dec5730744ddb1fac/lib/file_monitoring/file_monitoring.rb#L37)
- [iterates on instances](https://github.com/bbfsdev/bbfs/blob/8f853bb8c481d04e8e57e76dec5730744ddb1fac/lib/file_monitoring/file_monitoring.rb#L45)
- [adds file1 and file2 to monitoring under /dir1/dir2/](https://github.com/bbfsdev/bbfs/blob/8f853bb8c481d04e8e57e76dec5730744ddb1fac/lib/file_monitoring/monitor_path.rb#L164) (lines 164-171) (only two calls for load_instances at all (?))

Whether and when we add sub-dirs dir3, dir4?
If not then if new file  /dir1/dir2/dir3/file3 will be added, whether it will be cached by monitoring system?

genadyp:
@yarondbb 
suggest to add testcase

lib/file_monitoring/monitor_path.rb
yarondbb, line 185: for each instance we call the root tree dir ( /dir1/dir2/) to travel the tree and add the instance.
In this process, we create sub dirs:
Tree starts with 1 DirStat holding: dir1/dir2
when file1 is loaded to tree:
  dir1/dir2/dir3 is created under DirStat (which holds dir1/dir2)
  dir1/dir2/dir3/file1 is created as FileStat under DirStat (which holds dir1/dir2/dir3)
when file2 is loaded to tree:
Travel starts at DirStat (which holds dir1/dir2):
dir1/dir2/dir3 already exists so it continues with recursive call to it with load_instance
dir1/dir2/dir3/file4 is created under DirStat (which holds dir1/dir2/dir3)
dir1/dir2/dir3/dir4/file1 is created as FileStat under DirStat (which holds dir1/dir2/dir3/dir4)

genadyp:
```
for each instance we call the root tree dir ( /dir1/dir2/) to travel the tree and add the instance.
In this process, we create sub dirs:
Tree starts with 1 DirStat holding: dir1/dir2
when file1 is loaded to tree:
dir1/dir2/dir3 is created under DirStat (which holds dir1/dir2)
dir1/dir2/dir3/file1 is created as FileStat under DirStat (which holds dir1/dir2/dir3)
when file2 is loaded to tree:
Travel starts at DirStat (which holds dir1/dir2):
dir1/dir2/dir3 already exists so it continues with recursive call to it with load_instance
dir1/dir2/dir3/file4 is created under DirStat (which holds dir1/dir2/dir3)
dir1/dir2/dir3/dir4/file1 is created as FileStat under DirStat (which holds dir1/dir2/dir3/dir4)
```

could you point in the code, please, where exactly /dir1/dir2/dir3/ sub-directory will be added under DirStat rooted  on /dir1/dir2/
?

yarondbb:
it will happen in monitor_path.rb line 179

genadyp:
I know, but i am sorry i can't see, when it can happen in the mentioned example. This [else condition](https://github.com/bbfsdev/bbfs/blob/d9104ee9ca6f86d9fcc83101ff9306dd4f535247/lib/file_monitoring/monitor_path.rb#L174) looks like something we never reach, cause it looks like currently load_instance called only for files ([1](https://github.com/bbfsdev/bbfs/blob/d9104ee9ca6f86d9fcc83101ff9306dd4f535247/lib/file_monitoring/file_monitoring.rb#L45), [2](https://github.com/bbfsdev/bbfs/blob/d9104ee9ca6f86d9fcc83101ff9306dd4f535247/lib/file_monitoring/file_monitoring.rb#L71)). Looks like recursive call for load_instance absent while we are in the [file case](https://github.com/bbfsdev/bbfs/blob/d9104ee9ca6f86d9fcc83101ff9306dd4f535247/lib/file_monitoring/monitor_path.rb#L167).

yarondbb:
File is a leaf, so recursive call should not happen
Even though we want to load a file (say file1), the first call is to the root DirStat which holds the monitoring path dir1/dir2 with sub_paths_index of 2 (path of next dir to create: dir1/dir2/dir3)

Let's talk

genadyp:
found hole in my understanding.
no more questions from my side.

Issue #217, opened 2013-08-12
hanlde signals


yarondbb:
hanlde signals in 3 main cases:
1. terminaing the program - only critical actions and exit - need to define
1. Shutdown - close the application with out loosing data etc. - need to define
2. user signals - tell the application to perform specific tasks such as dump internal state etc.. - need to define

genadyp:
[example of signal handling](https://github.com/bbfsdev/bbfs/blob/master/lib/process_monitoring/monitoring_info.rb#L12)

PR #220, opened 2013-08-15
fixed. Root cause was endless while loop when execution finished with no...

yarondbb:
... exception and counter was not reduced below zero.

I cenceld the endless loop mechanism since we are not using it

### DIFF
--- a/bin/backup_server
+++ b/bin/backup_server
@@ -20,8 +20,6 @@ ContentServer.init_globals
 
 Thread.abort_on_exception = true  # TODO (genadyp) should be treated globally? by param for example.
 
-retries = 0  # number of retries to run a server
-
 begin
   RunInBackground.run { ContentServer.run_backup_server }
 rescue SystemExit, SignalException => exception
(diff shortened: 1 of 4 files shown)


bin/backup_server
```diff
@@ -20,8 +20,6 @@ ContentServer.init_globals
 
 Thread.abort_on_exception = true  # TODO (genadyp) should be treated globally? by param for example.
 
-retries = 0  # number of retries to run a server
-
 begin
   RunInBackground.run { ContentServer.run_backup_server }
 rescue SystemExit, SignalException => exception
```
genadyp, line 25: code duplication:
lines 27-29, 31-33

Consider remove.
Can it be connected to signal handling cause then, may be we indeed need conditions for various exceptions (abnormal termination, stop signal sent by user, immediate abort sent by user,...)?  

The same is due for other executable files.

yarondbb, line 25: This is part of Alexy task #217 
Closing since Genday performed the code review

PR #222, opened 2013-08-19
changed logs

yarondbb:


lib/log.rb
```diff
@@ -141,34 +141,38 @@ def Log.error(msg)
   end
 
   # Log info massages
-  def Log.info(msg)
+  def Log.info(msg, *args)
     Log.init if @log4r.nil?
+    msg = msg % args
```
genadyp, line 154: args elements should have overridden to_s method, otherwise we can get undesirable log output.

```
1.9.3-p327 :049 >   class CL
1.9.3-p327 :050?>   def initialize(str)
1.9.3-p327 :051?>     @param = str
1.9.3-p327 :052?>     end
1.9.3-p327 :053?>   end

1.9.3-p327 :060 >   def info(msg, *args)
1.9.3-p327 :061?>   msg = msg % args
1.9.3-p327 :062?>   end

1.9.3-p327 :065 >   john = CL.new("john")
 => #<CL:0x000000014c1660 @param="john"> 

1.9.3-p327 :067 > msg = "Hello %s %s"
 => "Hello %s %s" 

1.9.3-p327 :068 > info(msg, john, "Mary")
 => "Hello #<CL:0x000000014c1660> Mary" 

1.9.3-p327 :072 >   class CL1
1.9.3-p327 :073?>   def initialize(str)
1.9.3-p327 :074?>     @param = str
1.9.3-p327 :075?>     end
1.9.3-p327 :076?>   def to_s
1.9.3-p327 :077?>     @param.to_s
1.9.3-p327 :078?>     end
1.9.3-p327 :079?>   end
 => nil 

1.9.3-p327 :080 > john = CL1.new("john")
 => john 

1.9.3-p327 :081 > msg = "Hello %s %s"
 => "Hello %s %s" 

1.9.3-p327 :082 > info(msg, john, "Mary")
 => "Hello john Mary" 
```

lib/log.rb
```diff
@@ -141,34 +141,38 @@ def Log.error(msg)
   end
 
   # Log info massages
-  def Log.info(msg)
+  def Log.info(msg, *args)
```
genadyp, line 152: Need comment describing method parameters, especially first parameter that is a [format string](http://www.ruby-doc.org/core-1.9.3/String.html#method-i-25).
Need to update [example](https://github.com/bbfsdev/bbfs/blob/master/examples/log_example.rb)
Need to notify all the team regarding new usage, cause it's an important feature that we use a lot.

lib/log.rb
yarondbb, line 154: Great! We should take care that the loged objects have to_s

yarondbb, line 154: Also, some cases (such as net objects) we do want ruby to give us the inspection (no to_s) when we log it (ususally in debug level for us only)

PR #224, opened 2013-08-20
Update server.rb

AlexeyNemytov:
https://github.com/bbfsdev/bbfs/issues/217
Block force write to file


yarondbb:
good quick solution!

PR #231, opened 2013-09-11
Performence refactor

yarondbb:
Memory refactor:

Problem: 
Memory kept rising during Monitoring\Indexing\Write to file

Investigation: 
Ruby used a lot of temporary objects which made the internal allocated heaps arrays to grow. Not all temporary heaps used while monitoring\indexing\writing to file where freed. Probalby since there are also some living objects on the arrays which prevent to release it.

Solution:
  write a code which is much more memory fit so that minimum temporary objects are created.
     a. Seperate monitoring from Indexing. Doing these 2 tasks in paralel uses more memory and more time.
     b. Replace all Each loops with Enumerator loops. Each loops create temporary arrays of input for each  iteration. This can cause huge memory load
     c. Synchronize monitoring and write to file - after each monitor cycle. If ContentData canged, write to file. (this preent writing in paralel which increse memory load). Keeping also legacy, write to file, thread.
     d.  Write to file in chunks and Force GC in between
     e.  Force GC after each dir is monitored
     f. Force GC after each dir is indexed


yarondbb:
@genadyp it is all yours :)

yarondbb:
Using this version with extreem conditions (150K files. Monitor\index\write to file all the time) has process memory of 195M stable over days

lib/content_data/content_data.rb
```diff
@@ -84,9 +86,14 @@ def each_content(&block)
     # block is provided with: checksum, size, content modification time,
     #   instance modification time, server and file path
     def each_instance(&block)
-      @contents_info.keys.each { |checksum|
+      Log.info("inside each_instance: @contents_info:\n#{@contents_info.size}")
```
genadyp: Should it be Log.debug ?

yarondbb: I will removed it (was there for debuging)

lib/content_data/content_data.rb
```diff
@@ -84,9 +86,13 @@ def each_content(&block)
     # block is provided with: checksum, size, content modification time,
     #   instance modification time, server and file path
     def each_instance(&block)
-      @contents_info.keys.each { |checksum|
+      contents_enum = @contents_info.each_key
+      loop {
+        checksum = contents_enum.next rescue break
+        #Log.info("checksum:#{checksum}")
         content_info = @contents_info[checksum]
         content_info[1].keys.each {|location|
```
genadyp: Do we need here also replace iterator with enumerator?

lib/content_data/content_data.rb
```diff
@@ -94,6 +100,7 @@ def each_instance(&block)
                      location[0], location[1])
         }
       }
+      Log.info('end of each_instance')
```
genadyp: Consider replace with Log.debug.

lib/content_data/content_data.rb
yarondbb: usually we are not expecting to have many instances of the same content (1-10 or something of that kind), so I think this loop does not create to many temporary objects.
But worth to concider using iterator also here

lib/content_data/content_data.rb
```diff
@@ -252,14 +266,48 @@ def to_file(filename)
       FileUtils.makedirs(content_data_dir) unless File.directory?(content_data_dir)
       file = File.open(filename, 'w')
       file.write("#{@contents_info.length}\n")
-      each_content { |checksum, size, content_mod_time|
-        file.write("#{checksum},#{size},#{content_mod_time}\n")
+      contents_enum = @contents_info.each_key
+      content_chunks = @contents_info.length / 5000 + 1
```
genadyp, line 281: For better maintainability suggest to add a method comment describing why chunks used here cause it not obvious.

lib/content_data/content_data.rb
```diff
@@ -252,14 +266,48 @@ def to_file(filename)
...
+        GC.start
       }
       file.write("#{@instances_info.length}\n")
-      each_instance { |checksum, size, _, instance_mod_time, server, path|
-        file.write("#{checksum},#{size},#{server},#{path},#{instance_mod_time}\n")
+      contents_enum = @contents_info.each_key
+      instances_chunks = @instances_info.length / 5000 + 1
+      instances_chunks.times { |index|
+        to_file_instances_chunk(file,contents_enum,5000)
+        GC.start
```
genadyp, line 291: Is there any penalty for running GC with high frequency?

lib/content_server/server.rb
```diff
@@ -88,5 +89,24 @@ def monitor_general_process_vars
...
+    Log.info('Start flush local content data to file.')
+    $testing_memory_log.info('Start flush content data to file') if $testing_memory_active
+
+    $local_content_data_lock.synchronize{
+      local_content_data_unique_id = $local_content_data.unique_id
+      if (local_content_data_unique_id != $last_content_data_id)
+        $last_content_data_id = local_content_data_unique_id
+        $local_content_data.to_file($tmp_content_data_file)
+        File.rename($tmp_content_data_file, Params['local_content_data_path'])
+        Log.info('End flush local content data to file.')
```
genadyp: Consider Log.debug instead Log.info

lib/content_server/server.rb
```diff
@@ -88,5 +89,24 @@ def monitor_general_process_vars
...
+        $testing_memory_log.info('End flush content data to file') if $testing_memory_active
+      else
+        Log.info('no need to flush. content data has not changed')
```
genadyp: Consider Log.debug instead Log.info

lib/content_server/server.rb
```diff
@@ -88,5 +89,24 @@ def monitor_general_process_vars
     end
   end
 
-  module_function :init_globals, :handle_program_termination, :monitor_general_process_vars
+  def flush_content_data
+    Log.info('Start flush local content data to file.')
```
genadyp: Consider Log.debug instead Log.info

lib/file_monitoring/monitor_path.rb
```diff
@@ -21,76 +19,78 @@ class FileStatEnum
...
-        @size = file_stats.size
-        #@creation_time = file_stats.ctime.utc
-        @modification_time = file_stats.mtime.to_i
-        @cycles = 0
-      else
-        new_state = FileStatEnum::UNCHANGED
-        @cycles += 1
-        if @cycles >= @stable_state
-          new_state = FileStatEnum::STABLE
+    def index
```
genadyp, line 73: Do we have any use of [FileIndexing](https://github.com/bbfsdev/bbfs/tree/performence_refactor/lib/file_indexing)?
If no then consider removing FileIndexing module.

genadyp:
Good work. Thank you!
Few general comments:
- Usage of Enumerators instead of Iterators should be documented somewhere and should be clear for all developers
- Consider summarizing your performance findings in document. It very important for maintainability and preventing such a problems in the future.
- Consider adding tests that cover your code
- Run and if needed fix tests after the merge

lib/content_data/content_data.rb
yarondbb, line 291: Yes. GC takes time.

kolmanv, line 291: We can add test to check no one uses iterators but rather only
enumerators... this should not be too hard..

lib/content_data/content_data.rb
yarondbb, line 281: done

lib/content_server/server.rb
yarondbb: done

lib/content_server/server.rb
yarondbb: done

lib/content_server/server.rb
yarondbb: done

lib/file_monitoring/monitor_path.rb
yarondbb, line 73: Good point.
Currently only auxiliary methods are used: get_checksum and get_content_checksum
We could move them to server.rb or something..
@kolmanv, what do you think ?

bbfsdev, line 73: I think removing is good idea. Please clean this code and don't leave any unused code. Also delete the whole [legacy](https://github.com/bbfsdev/bbfs/tree/master/lib/legacy) directory

PR #242, opened 2014-01-08
Issue 237 monitor dir rename

yarondbb:
Summary of solution for this branch:
User flow:
1. Stop the application
2. Run Do manual changes in monitoring directories (any change)
3. Run the application with new flag: manual_file_change=true
4. Application should provide new content data file with the manual changes (without doing any indexing), and exit
5. Run application again with new flag: manual_file_change=false (default calue). Application will load the uchanged content data file and continue

Alogorithm for step 3:
  load content data with new flag: 'manual_file_change'
  During load the application will builld new map:
    file,size,mod --> [checksum, uniqu]
  If key found more then 1 time,then uniqu become false. This will indicate not to do anything with it during manual change

  monitoring new behavior:
    existing files in Application Tree - ignore
    new found files in OS system:
      If not exist in map - need to be indexed. skip (will be handled in step 5.
      If exist in map (copied or moved):
        if uniqu is true - take the original checksum and add to ContentData
        if uniqu is false - need to be indexed. skip. will be handled in step 5.
    unmarked files (non existing) - will have legacy behavior and be remove from content data
  Write new content data to file
  Exit the application

### DIFF
--- a/lib/content_server/server.rb
+++ b/lib/content_server/server.rb
@@ -32,6 +32,7 @@ module ContentServer
     $remote_content_data_lock = nil
     $remote_content_data = nil
     $last_content_data_id = nil
+    $file_attr_to_ref = {}
   end
 
   def handle_program_termination(exception)
--- a/lib/content_server/server.rb
+++ b/lib/content_server/server.rb
@@ -32,7 +32,7 @@ module ContentServer
     $remote_content_data_lock = nil
     $remote_content_data = nil
     $last_content_data_id = nil
-    $file_attr_to_ref = {}
+    $file_attr_to_checksum = {}  # used for monitor manual file changes feature
   end
 
   def handle_program_termination(exception)
--- a/lib/content_server/server.rb
+++ b/lib/content_server/server.rb
@@ -32,7 +32,6 @@ module ContentServer
     $remote_content_data_lock = nil
     $remote_content_data = nil
     $last_content_data_id = nil
-    $file_attr_to_checksum = {}  # used for monitor manual file changes feature
   end
 
   def handle_program_termination(exception)
--- a/lib/file_monitoring.rb
+++ b/lib/file_monitoring.rb
@@ -20,8 +20,8 @@ module FileMonitoring
               'This log containd track of changes found during monitoring')
   Params.complex('monitoring_paths', [], 'Array of Hashes with 3 fields: ' \
                  'path, scan_period and stable_state.')
-  Params.boolean('manual_file_changes', false, 'true, indicates to application to check if new files were copied ' \
-                 'or moved from other location. In this case content data file should be updated accordingly')
+  Params.boolean('manual_file_changes', false, 'true, indicates to application that a set of files were ' \
+    ' moved/copied we we should not index them but rather copy their hashes from contents there were copied from')
 
   # @see FileMonitoring#monitor_files
   def monitor_files
--- a/lib/file_monitoring.rb
+++ b/lib/file_monitoring.rb
@@ -20,6 +20,8 @@ module FileMonitoring
               'This log containd track of changes found during monitoring')
   Params.complex('monitoring_paths', [], 'Array of Hashes with 3 fields: ' \
                  'path, scan_period and stable_state.')
+  Params.boolean('manual_file_changes', false, 'true, indicates to application to check if new files were copied ' \
+                 'or moved from other location. In this case content data file should be updated accordingly')
 
   # @see FileMonitoring#monitor_files
   def monitor_files
--- a/lib/file_monitoring/file_monitoring.rb
+++ b/lib/file_monitoring/file_monitoring.rb
@@ -69,9 +69,11 @@ module FileMonitoring
             unless ident_file_info
               #  Add file checksum to map
               $file_attr_to_checksum[file_attr_str] = IdentFileInfo.new(checksum)
+              puts "Added file:#{file_attr_str} to map"
             else
               # File already in map. Need to mark as not unique
               ident_file_info.unique = false  # file will be skipped if found at new location
+              puts "File #{file_attr_str} already exists"
             end
           end
           # construct sub paths array from full file path:
--- a/lib/file_monitoring/file_monitoring.rb
+++ b/lib/file_monitoring/file_monitoring.rb
@@ -112,22 +112,24 @@ module FileMonitoring
           # -------------------------- MANUAL MODE
           # ------------ LOOP DIRS
           dir_stat_array.each { | dir_stat|
-            Log.info("In Manual mode. Start monitor path:%s. moved or copied files (same name, size and time " +
-                         "modification) will use the checksum of the original files and be updated in " +
-                         "content data file", dir_stat[0].path)
-            $testing_memory_log.info("Start monitor path:#{dir_stat[0].path} moved or copied files (same name, size and " +
-                                         "time modification) will use the checksum of the original files and be updated in " +
-                                         "content data file") if $testing_memory_active
+            log_msg = "In Manual mode. Start monitor path:%s. moved or copied files (same name, size and time " +
+                "modification) will use the checksum of the original files and be updated in " +
+                "content data file" % [dir_stat[0].path]
+            Log.info(log_msg)
+            $testing_memory_log.info(log_msg) if $testing_memory_active
+
             # ------- MONITOR
             dir_stat[0].monitor(file_attr_to_checksum)
 
             # ------- REMOVE PATHS
             # remove non existing (not marked) files\dirs
-            Log.info('Start remove non existing paths')
-            $testing_memory_log.info('Start remove non existing paths') if $testing_memory_active
+            log_msg = 'Start remove non existing paths'
+            Log.info(log_msg)
+            $testing_memory_log.info(log_msg) if $testing_memory_active
             dir_stat[0].removed_unmarked_paths
-            Log.info('End monitor path and index')
-            $testing_memory_log.info('End monitor path and index') if $testing_memory_active
+            log_msg = 'End monitor path and index'
+            Log.info(log_msg)
+            $testing_memory_log.info(log_msg) if $testing_memory_active
           }
 
           # ------ WRITE CONTENT DATA
--- a/lib/file_monitoring/file_monitoring.rb
+++ b/lib/file_monitoring/file_monitoring.rb
@@ -57,6 +57,7 @@ module FileMonitoring
       #Look over loaded content data if not empty
       # If file is under monitoring path - Add to DirStat tree as stable with path,size,mod_time read from file
       # If file is NOT under monitoring path - skip (not a valid usage)
+      file_attr_to_checksum = {}  # This structure is used to optimize indexing when user specifies a directory was moved.
       unless $local_content_data.empty?
         Log.info("Start build data base from loaded file. This could take several minutes")
         inst_count = 0
@@ -64,11 +65,11 @@ module FileMonitoring
             |checksum, size, _, mod_time, _, path, index_time|
 
           if Params['manual_file_changes']
-            file_attr_str = File.basename(path) + size.to_s + mod_time.to_s
-            ident_file_info = $file_attr_to_checksum[file_attr_str]
+            file_attr_key = [File.basename(path), size, mod_time]
+            ident_file_info = file_attr_to_checksum[file_attr_key]
             unless ident_file_info
               #  Add file checksum to map
-              $file_attr_to_checksum[file_attr_str] = IdentFileInfo.new(checksum, index_time)
+              file_attr_to_checksum[file_attr_key] = IdentFileInfo.new(checksum, index_time)
             else
               # File already in map. Need to mark as not unique
               ident_file_info.unique = false  # file will be skipped if found at new location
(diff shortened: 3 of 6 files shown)


lib/file_monitoring.rb
```diff
@@ -20,6 +20,8 @@ module FileMonitoring
               'This log containd track of changes found during monitoring')
   Params.complex('monitoring_paths', [], 'Array of Hashes with 3 fields: ' \
                  'path, scan_period and stable_state.')
+  Params.boolean('manual_file_changes', false, 'true, indicates to application to check if new files were copied ' \
+                 'or moved from other location. In this case content data file should be updated accordingly')
```
bbfsdev: The parameter explanation is not exact.
true, indicates to application that a set of files were moved/copied we we should not index them but rather copy their hashes from contents there were copied from.

lib/content_server/server.rb
```diff
@@ -32,6 +32,7 @@ def init_globals
     $remote_content_data_lock = nil
     $remote_content_data = nil
     $last_content_data_id = nil
+    $file_attr_to_checksum = {}  # used for monitor manual file changes feature
```
bbfsdev: The comment does not say enough.
This structure is used to optimize indexing when user specifies a directory was moved.

lib/file_monitoring/file_monitoring.rb
```diff
@@ -23,7 +23,29 @@ class FileMonitoring
     # This methods controlled by <tt>monitoring_paths</tt> configuration parameter,
     # that provides path and file monitoring configuration data
     def monitor_files
-      conf_array = Params['monitoring_paths']
+      
+      #init log4r
```
bbfsdev, line 27: You just copied this part from below, right?

lib/file_monitoring/file_monitoring.rb
```diff
@@ -39,7 +61,19 @@ def monitor_files
         Log.info("Start build data base from loaded file. This could take several minutes")
         inst_count = 0
         $local_content_data.each_instance {
-            |_, size, _, mod_time, _, path|
+            |checksum, size, _, mod_time, _, path, index_time|
+
+          if Params['manual_file_changes']
+            file_attr_str = File.basename(path) + size.to_s + mod_time.to_s
```
bbfsdev: Don't use str as key, use array.

Think about this case:
name:file11
size:22
mod_time:33

name: file
size:112
mod_time:233

You can use the following as key (don't even change to string! numbers take less memory)
key = [File.basename(path), size, mod_time]

lib/file_monitoring/monitor_path.rb
```diff
@@ -19,6 +19,18 @@ class FileStatEnum
     STABLE = "STABLE"
   end
 
+  # Used for dir rename. Holds following info:
+  #   ref_counter = number of references to same file which exist under several DirStat
```
bbfsdev: This comment is not for the IdentFileInfo class.
Remove or update/change.

lib/file_monitoring/monitor_path.rb
```diff
@@ -297,60 +310,90 @@ def monitor
...
           else
-            # new File child:
-            child_stat = FileStat.new(globed_path, FileStatEnum::NEW,
-                                      globed_path_stat.size, globed_path_stat.mtime.to_i)
-            @@log.info("NEW file: " + globed_path)
-            @@log.outputters[0].flush if Params['log_flush_each_message']
-            child_stat.marked = true
-            add_file(child_stat)
+            # ---------------------------- NEW FILE ----------
+            unless Params['manual_file_changes']
```
bbfsdev, line 367: unless is not readable as if, consider changing to if and replace with the else block.

lib/content_server/server.rb
bbfsdev: Don't use global variable, it is not really needed.
You only need it once in the monitor, create it there and propagate where needed.

lib/file_monitoring/file_monitoring.rb
```diff
@@ -72,6 +106,40 @@ def monitor_files
...
         $last_content_data_id = $local_content_data.unique_id
+
+        if Params['manual_file_changes']
+          # -------------------------- MANUAL MODE
+          # ------------ LOOP DIRS
+          dir_stat_array.each { | dir_stat|
+            Log.info("In Manual mode. Start monitor path:%s. moved or copied files (same name, size and time " +
+                         "modification) will use the checksum of the original files and be updated in " +
+                         "content data file", dir_stat[0].path)
+            $testing_memory_log.info("Start monitor path:#{dir_stat[0].path} moved or copied files (same name, size and " +
```
bbfsdev: What is testing memory log? Why is it needed?
It has the same string to print?

lib/file_monitoring/monitor_path.rb
```diff
@@ -297,60 +310,90 @@ def monitor
...
+                    @@log.outputters[0].flush if Params['log_flush_each_message']
+                  else
+                    @@log.info("UNCHANGED file: " + globed_path)
+                    @@log.outputters[0].flush if Params['log_flush_each_message']
+                  end
                 end
               end
+            else
+              # --------- MANUAL MODE
+              child_stat.marked = true
```
bbfsdev, line 363: Why there is nothing to do in case of manual file changes?
Please replace the unless with if will be more readable.

lib/file_monitoring/monitor_path.rb
```diff
@@ -297,60 +310,90 @@ def monitor
...
+                                        globed_path_stat.size,
+                                        globed_path_stat.mtime.to_i)
+              @@log.info("NEW file: " + globed_path)
+              @@log.outputters[0].flush if Params['log_flush_each_message']
+              child_stat.marked = true
+              add_file(child_stat)
+            else
+              # --------------------- MANUAL MODE
+              # check if file name and attributes exist in global file attr map
+              file_attr_str = File.basename(globed_path) + globed_path_stat.size.to_s + globed_path_stat.mtime.to_i.to_s
```
bbfsdev: Change the key to array.

lib/file_monitoring/monitor_path.rb
```diff
@@ -297,60 +310,90 @@ def monitor
...
+              file_ident_info = $file_attr_to_checksum[file_attr_str]
+              # If not found (real new file) or found but not unique then file needs indexing. skip in manual mode.
+
```
bbfsdev: remove this line as the comment refers to unless.

lib/file_monitoring/monitor_path.rb
```diff
@@ -297,60 +310,90 @@ def monitor
         # Get File \ Dir status
         globed_path_stat = File.lstat(globed_path) rescue next  # File or dir removed from OS file system
         if globed_path_stat.file?
-          # File case
+          # ----------------------------- FILE -----------------------
```
bbfsdev, line 320: Break this function! This is too long and hard to understand. Make it as several functions. It should be easy to read and understand too much nested if and for.

bbfsdev:
Good job!

yarondbb:
@kolmanv, note that Genady already reviewd file: lib/content_data/content_data.rb (in issue #243).
I merged 243 branch to this branch, since 'adding index time' feature is also part of this feature (monitoring manual changes)

lib/file_monitoring/file_monitoring.rb
yarondbb, line 27: yes

lib/file_monitoring/file_monitoring.rb
yarondbb: Ok. good catch of bug.

lib/file_monitoring/file_monitoring.rb
yarondbb: It is the logger of the memory test application. Here it takes the same log. In other places it produces its own memory log.

lib/file_monitoring/monitor_path.rb
yarondbb, line 320: I find it very difiicult to trace an algorithm where you have many inner funciotns. In our case, 1 of them should perform a recursive call making the Tree of calls complicated.
Also, for example, when you look at some inner function, then you need to see where did the parameters come from, you need to find the caller.. it takes time.. parameters change their names.. etc..
For me, 1 script code, is much more easy to follow. 
Why ? It is all in 1 place. 
All parameter names stay this way. 
Upper lines numbers mean 'before' in algorithm flow. Lowwer lines mean 'after' in the algorithm flow.
To ease orientation I use '-----------------------' for seperating major parts of the algorithm.
Maybe we could have a team discussion over this some time ?

lib/file_monitoring/monitor_path.rb
yarondbb, line 363: If we already have the file file in the Tree data base it means it was loaded already from content data file.
The only files we need to hanlde are the new files which were moved\copied from other location (which exists in the content data file)

My openion:
The unless is above since this is the common case (the manual flag will be used rarly). 
In terms of performence, the jump prediction of if else clauses takes the first block as the true prediction.
I think that the comment MANUAL MODE  is good to understand the block. 
What do you think?

lib/file_monitoring/monitor_path.rb
yarondbb: right, comment refere to existing lines or next line (unless line). Or I am missing something ?

lib/file_monitoring/file_monitoring.rb
bbfsdev: consider creating a string and passing it to both functions.

lib/file_monitoring/monitor_path.rb
bbfsdev, line 320: The problem is the spaces, I don't know what if i'm looking into where it starts and where it ends. Its a mess.

```
def monitor

  loop do
    next if is_symlink() || non_valid_utf8()

    if file
      handle_file
    else
      handle_dir
    end
  end // loop
  GC.start
end
```

handle file should also be split:

```
def handle_file(...)
  if alredy_exists
    handle_existing_file
  else
    handle_new_file
  end
end
```

lib/file_monitoring/monitor_path.rb
bbfsdev, line 363: I think that if else is 3 lines, I don't want to lookup what if it is (too long up) to understand the code.
If those 3 lines would be right below the if, it would be easy to understand what it relates to.
Other longer code will follow right away.

lib/file_monitoring/monitor_path.rb
yarondbb, line 320: I find it very difiicult to trace an algorithm where you have many inner funciotns. In our case, 1 of them should perform a recursive call making the Tree of calls complicated.
Also, for example, when you look at some inner function, then you need to see where did the parameters come from, you need to find the caller.. it takes time.. parameters change their names.. etc..
For me, 1 script code, is much more easy to follow. 
Why ? It is all in 1 place. 
All parameter names stay this way. 
Upper lines numbers mean 'before' in algorithm flow. Lowwer lines mean 'after' in the algorithm flow.
To ease orientation I use '-----------------------' for seperating major parts of the algorithm.
Maybe we could have a team discussion over this some time ?

lib/file_monitoring/monitor_path.rb
yarondbb, line 363: We can add documentation this way:
else  # case Params['manual_file_changes']

lib/file_monitoring/monitor_path.rb
yarondbb, line 367: Discussion on same issue is above

lib/file_monitoring/file_monitoring.rb
yarondbb: ok

lib/file_monitoring/monitor_path.rb
yarondbb, line 320: The method has this documentation:

# Algorithm:

```
  # assume that current dir is present
  # ls (glob) the dir path for child dirs and files
  # if child file is not already present, add it as new, mark it and handle its state
  # if file already present, mark it and handle its state.
  # if child dir is not already present, add it as new, mark it and propagates
  #    the recursive call
  # if child dir already present, mark it and handle its state
  # marked files will not be remove in next remove phase
```

We could exmplain better the pseudo algorithm there..

yarondbb, line 320: Note that this discussion is not related to the feature we are pushing to this release..

lib/file_monitoring/monitor_path.rb
bbfsdev, line 363: sounds good, also for 

```
end  # case Params['manual_file_changes']
```

two spaces after `end`.

lib/file_monitoring/monitor_path.rb
bbfsdev, line 320: I think it is related. Lets talk about this today at meeting.
I think we should write self explainable code. Simple and a programmer should not look at more then 10 lines each time.
We should think in closed blocks which are easy to test. How you suggest testing this one big function?
If each block is well testing then everything is easier.

You also mentioned you want to add two tests for this method. I think this should be done in this release.
I can share the effort.

yarondbb, line 320: The method was not written in this feature. It was there for more then a month.
Why pushing the release in time? There are many refactoring could be done in our code.
I suggest having this discussion off-line to this feature

yarondbb, line 320: Testing could be done on this method as a bleck box which should provide correct monitoring output for files,dirs,..
For me multiple functions are causing spagetty code which is not clear and hard to follow.
At my work we moved to big methods code rather then splitting to micro function code

lib/file_monitoring/monitor_path.rb
yarondbb, line 363: ok

Issue #235, opened 2013-12-17
Bug-found at release 1.4.0-bug-1


yarondbb:
The new version does not run with "enable_monitoring: true".
So I have run the "backup_server" without monitoring.

The next error was triggered with "enable_monitoring: true":
[bbfs@swift1 ~]$ backup_server                                                                                                                                                

Interrupt or Exit happened in server:''.
Exception type:'NoMethodError'.
Exception message:'undefined method `synchronize' for nil:NilClass'.
Stopping process.
Backtrace:
/usr/local/rvm/gems/ruby-1.9.3-p429@bbfs/gems/content_server-1.4.0/lib/networking/tcp.rb:197:in`block (2 levels) in start_reading'
/usr/local/rvm/gems/ruby-1.9.3-p429@bbfs/gems/content_server-1.4.0/lib/networking/tcp.rb:191:in `loop'
/usr/local/rvm/gems/ruby-1.9.3-p429@bbfs/gems/content_server-1.4.0/lib/networking/tcp.rb:191:in`block in start_reading'

yarondbb:
Root cause found:
At lib/networking/tcp.rb:197: @remote_server_available_mutex is used in method:start_reading.
Method start_reading is called (in initialize) before @remote_server_available_mutex is defined

bbfsdev:
IIUC this is the second time we fail on this issue. Please write a test and send for review.

PR #253, opened 2014-01-25
Fix lock initialization

bbfsdev:
I could not recreate the failure in #235 and therefore cannot write test for this.
Please check code.

@yarondbb 
@genadyp


yarondbb:
Looks good pls merge

Issue #236, opened 2013-12-17
Bug-found at release 1.4.0-handle symbolic links


yarondbb:
Shiray: You must not process them as a regular files or directories.
The symbolic point must be stored as is if it points to file/directory within "indexing area" or ignored in other case.

yarondbb:
work around released in Patch 1.4.1.
Work around is to ignore symbolic link files during Monitoring.

bbfsdev:
Yaron, please add a test for this case. Send it for review.

yarondbb:
tests written

yarondbb:
#254 submitted
Who can do the code review? @kolmanv  @genadyp  ?

kolmanv:
I will do this.

PR #254, opened 2014-03-01
Added symlink support.

yarondbb:
Added symlink support:
1. Monitoring symlinks
2. Loading symlinks from content data file

New content data file format:
  number of contents
  content1
  content2
  ..
  number of instances
  instance1
  instance2
  ..
  number of symlinks
  symlink1
  symlink2
  ..

Where symlink format is:
  server, symlink path, symlink target

### DIFF
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -46,7 +46,7 @@ module ContentData
     # Content Data unique identification
     # @return [ID] hash identification
     def unique_id
-      [@instances_info.hash,@symlinks_info.hash]
+      [@contents_info.hash,@symlinks_info.hash]
     end
 
     def clone_instances_info
@@ -277,35 +277,8 @@ module ContentData
       }
     end
 
-
     def ==(other)
-      # general quick checks
-      return false if other.nil?
-      return false if @contents_info.length != other.contents_size
-      return false if @instances_info.length != other.instances_size
-      return false if @symlinks_info.length != other.symlinks_size
-
-      # check instances
-      other.each_instance { |checksum, size, content_mod_time, instance_mod_time, server, path|
-        return false if instance_exists(path, server) != other.instance_exists(path, server)
-        local_content_info = @contents_info[checksum]
-        return false if local_content_info.nil?
-        return false if local_content_info[0] != size
-        return false if local_content_info[2] != content_mod_time
-        #check instances
-        local_instances =  local_content_info[1]
-        return false if other.checksum_instances_size(checksum) != local_instances.length
-        location = [server, path]
-        local_instance_mod_time, _ = local_instances[location]
-        return false if local_instance_mod_time.nil?
-        return false if local_instance_mod_time != instance_mod_time
-      }
-
-      # check symlinks
-      other.each_symlink { |server, path, target|
-        return false if symlink_exists(path, server) != other.symlink_exists(path, server)
-      }
-      true
+      unique_id  == other.unique_id
     end
 
     def remove_content(checksum)
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -278,6 +278,7 @@ module ContentData
     end
 
     def ==(other)
+      return nil if other.nil?  # for this case: content_data == nil
       unique_id  == other.unique_id
     end
 
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -35,16 +35,18 @@ module ContentData
       if other.nil?
         @contents_info = {}  # Checksum --> [size, paths-->time(instance), time(content)]
         @instances_info = {}  # location --> checksum to optimize instances query
+        @symlinks_info = {}  # [server,symlink path] -> target
       else
         @contents_info = other.clone_contents_info
         @instances_info = other.clone_instances_info  # location --> checksum to optimize instances query
+        @symlinks_info =  other.clone_symlinks_info
       end
     end
 
     # Content Data unique identification
     # @return [ID] hash identification
     def unique_id
-      @instances_info.hash
+      [@instances_info.hash,@symlinks_info.hash]
     end
 
     def clone_instances_info
@@ -81,6 +83,16 @@ module ContentData
                               content_time]
       }
       clone_contents_info
+    end
+
+    def clone_symlinks_info
+      symlinks_info_enum = @symlinks_info.each_key
+      cloned_symlinks = {}
+      loop {
+        symlink_key = symlinks_info_enum.next rescue break
+        cloned_symlinks[[symlink_key[0].clone, symlink_key[0].clone]] = @symlinks_info[symlink_key].clone
+      }
+      cloned_symlinks
     end
 
     # iterator over @contents_info data structure (not including instances)
@@ -131,12 +143,27 @@ module ContentData
       }
     end
 
+    # iterator over @symlinks_info data structure
+    # block is provided with: server, file path and target
+    def each_symlink(&block)
+      symlink_enum = @symlinks_info.each_key
+      loop {
+        symlink_key = symlink_enum.next rescue break
+        symlink_target = @symlinks_info[symlink_key]
+        block.call(symlink_key[0], symlink_key[1], symlink_target)
+      }
+    end
+
     def contents_size()
       @contents_info.length
     end
 
     def instances_size()
       @instances_info.length
+    end
+
+    def symlinks_size()
+      @symlinks_info.length
     end
 
     def checksum_instances_size(checksum)
@@ -181,8 +208,16 @@ module ContentData
       @instances_info[location] = checksum
     end
 
+    def add_symlink(server, path, target)
+      @symlinks_info[[server,path]] = target
+    end
+
+    def remove_symlink(server, path)
+      @symlinks_info.delete([server,path])
+    end
+
     def empty?
-      @contents_info.empty?
+      @contents_info.empty? and @symlinks_info.empty?
     end
 
     def content_exists(checksum)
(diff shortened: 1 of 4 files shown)


yarondbb:
This is the long term solution for issue #236

lib/content_data/content_data.rb
```diff
@@ -35,16 +35,18 @@ def initialize(other = nil)
...
         @instances_info = other.clone_instances_info  # location --> checksum to optimize instances query
+        @symlinks_info =  other.clone_symlinks_info
       end
     end
 
     # Content Data unique identification
     # @return [ID] hash identification
     def unique_id
-      @instances_info.hash
+      [@instances_info.hash,@symlinks_info.hash]
```
bbfsdev: Why content info is not part of this unique id?

lib/content_data/content_data.rb
```diff
@@ -83,6 +85,16 @@ def clone_contents_info
       clone_contents_info
     end
 
+    def clone_symlinks_info
+      symlinks_info_enum = @symlinks_info.each_key
+      cloned_symlinks = {}
+      loop {
+        symlink_key = symlinks_info_enum.next rescue break
+        cloned_symlinks[[symlink_key[0].clone, symlink_key[0].clone]] = @symlinks_info[symlink_key].clone
```
bbfsdev, line 93: Why do you need to clone the values too?
Why don't you clone only the hash?
http://stackoverflow.com/questions/4157399/how-do-i-copy-a-hash-in-ruby

lib/content_data/content_data.rb
```diff
@@ -181,8 +208,16 @@ def add_instance(checksum, size, server, path, modification_time, index_time=Tim
...
+      @symlinks_info[[server,path]] = target
+    end
+
+    def remove_symlink(server, path)
+      @symlinks_info.delete([server,path])
+    end
+
     def empty?
-      @contents_info.empty?
+      @contents_info.empty? and @symlinks_info.empty?
```
bbfsdev, line 220: What about instances_info?

lib/content_data/content_data.rb
```diff
@@ -216,7 +256,9 @@ def remove_directory(dir_to_remove, server)
       loop {
         checksum = contents_enum.next rescue break
         instances =  @contents_info[checksum][1]
-        instances.each_key { |location|
```
bbfsdev, line 219: Please remind me why we use loops? Is this the memory issue?

lib/content_data/content_data.rb
```diff
@@ -224,12 +266,26 @@ def remove_directory(dir_to_remove, server)
         }
         @contents_info.delete(checksum) if instances.empty?
       }
+
+      # handle symlinks
+      symlinks_enum = @symlinks_info.each_key
+      loop {
+        symlink_key = symlinks_enum.next rescue break
+        if symlink_key[0] == server and symlink_key[1].scan(dir_to_remove).size > 0
+          @symlinks_info.delete(symlink_key)
```
bbfsdev, line 275: Can we delete items from hash while iterating over it?

lib/content_data/content_data.rb
```diff
@@ -244,6 +300,11 @@ def ==(other)
         return false if local_instance_mod_time.nil?
         return false if local_instance_mod_time != instance_mod_time
       }
+
+      # check symlinks
+      other.each_symlink { |server, path, target|
+        return false if symlink_exists(path, server) != other.symlink_exists(path, server)
```
bbfsdev: Why you check here only for existence and not for target too?

lib/content_data/content_data.rb
```diff
@@ -382,24 +456,51 @@ def from_file(filename)
...
+          raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                    "number of symlinks should be a Number. We got:#{number_of_symlinks}")
+        end
+        number_of_symlinks.to_i.times {
+          symlinks_line = file.gets
+          unless symlinks_line
+            raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                   "Expected to read symlink line but reached EOF")
+          end
+          parameters = symlinks_line.split(',')
```
bbfsdev, line 446: What is "," is part of file path?

We have to rewrite the serialization part!

lib/file_monitoring/file_monitoring.rb
```diff
@@ -105,7 +106,44 @@ def monitor_files
             break
           }
         }
-        Log.info("End build data base from loaded file. loaded instances:#{inst_count}")
+
```
bbfsdev: This symlink handling should be sub-method

lib/file_monitoring/file_monitoring.rb
```diff
@@ -105,7 +106,44 @@ def monitor_files
...
+            |_, symlink_path, symlink_target|
+
+          # construct sub paths array from symlink path:
+          # Example:
+          #   symlink path = /dir1/dir2/file_name
+          #   Sub path 1: /dir1
+          #   Sub path 2: /dir1/dir2
+          #   Sub path 3: /dir1/dir2/file_name
+          # sub paths should match the paths of DirStat objs in the tree to reach the symlink location in Tree.
+          split_path = symlink_path.split(File::SEPARATOR)
```
bbfsdev: 1) This part is recurring, so you should extract it to function.
2) Seems like much redundant work here. Why cannot you just search for similar prefix:
instead of:
`sub_paths_index = sub_paths.index(dir_stat[0].path)`
do:
`symlink_path.match("^" + dir_stat[0])`

If you want to get the next directory you can use dirname and basebname of
http://www.ruby-doc.org/core-2.1.1/File.html

spec/file_monitoring/monitor_path_spec.rb
```diff
@@ -28,14 +33,40 @@ module Spec
         File.stub(:symlink?).with('symlink_file').and_return(true)
         return_lstat = File.lstat(__FILE__)
         File.stub(:lstat).with(any_args).and_return(return_lstat)
+        File.stub(:readlink).with('symlink_file').and_return('symlink_target')
+        # Perform monitoring
+        @dir_stat.monitor
+
+        # check that 'regular_file' was added to @files map under DirStat
+        @dir_stat.instance_eval{@files}['regular_file'].nil?.should be(false)
```
bbfsdev: nil? should be false is hard to understand use:
`dir_stat.instance_variable_get(:@files).has_key?('const_file').should be(true)`

lib/file_monitoring/monitor_path.rb
```diff
@@ -413,6 +457,20 @@ def monitor(file_attr_to_checksum=nil)
           child_stat.monitor(file_attr_to_checksum)
         end
       end
+
+      # check if any symlink was removed and update current symlinks map
+      symlinks_enum = @symlinks.each_key
+      loop {
+        symlink_key = symlinks_enum.next rescue break
+        if !found_symlinks.has_key?(symlink_key)
```
bbfsdev, line 465: Is this part of the `marked` algorithm?

lib/content_data/content_data.rb
yarondbb: The assumption is that contents_info and instances_info has a 1:1 instance match. but contents_info has also contents data. I think better to use contents_info (which includes instances_info data in it) and symlinks_info like this:
[@contents_info.hash,@symlinks_info.hash]

lib/content_data/content_data.rb
yarondbb, line 93: To be on the safe side.. It may be redundant

lib/content_data/content_data.rb
yarondbb, line 220: instances_info is a helper data structure to map instnaces to checksums.
If contents_info is empty it means that instnaces_info must be empty too or the ContentData is corrupted

lib/content_data/content_data.rb
yarondbb, line 219: yes. I sppoted another code which can be refactored using enum instead of each_key loop

lib/content_data/content_data.rb
yarondbb, line 275: Yes, since we iterate with the enumerator

lib/content_data/content_data.rb
yarondbb: Good catch. Thinking of it.. we could use the method unique_id here..
  unique_id == other.unique_id

lib/content_data/content_data.rb
yarondbb, line 446: hmmmm bad case since we have 2 paths at same line.. 
Maybe we need to choose a different delimiter? like '<' This is ilegal char both in Win\Unix
This time we perform file format change(symlinks), we can concider changing the format

lib/file_monitoring/file_monitoring.rb
yarondbb: Adding sub method both for symlinks and for regular files

lib/file_monitoring/file_monitoring.rb
yarondbb: This will give us if dir_stat[0] is root of this symlink
We also need the index in sub_paths. So sub_paths.index gives us the 2 birds ;)

yarondbb: I moved to use DirName it is much better!

lib/file_monitoring/monitor_path.rb
yarondbb, line 465: No I did not used the mark algorithm here
Instead, each dir has a saved @symlinks map. Then it creates the new current map and acts on diffs
Mark is not needed since the Symlink map is very basic data structure

spec/file_monitoring/monitor_path_spec.rb
yarondbb: fixed

lib/file_monitoring/monitor_path.rb
yarondbb, line 465: Fixed all ises. Pls review again

Issue #246, opened 2014-01-14
Get a diff between old and new content data files


yarondbb:
Should be done offline on 2 files.
Generate reports, which contents Added, Removed from old to new

bbfsdev:
Please look [file_utils.rb L179](https://github.com/bbfsdev/bbfs/blob/master/lib/file_utils/file_utils.rb#L179), you can see there are 3 actions available for content data (merge, intersect and minus) those are enough to generate the diff.
merge - combine two CD (symmetric)
intersect - return only instances/contents that are in both CD (symmetric)
minus - returns contents/instances that are in A and not in B (asymmetric)

genadyp:
@yarondbb 
@AlexeyNemytov 
@kolmanv 
@vshiray 
Few questions, cause usage is unclear:
- Only diff of contents should be reported?
- What should be a report format - what is an information that user wants to retrieve?
  Say, if we have only content data in the report (checksum, time) - what user can to do with this data?

kolmanv:
First I think the output should be a file in the same format.
It probably should be asymmetric, i.e., give me all instances of contents
which are in content a and are not in content b.
You should output something like:

bbfs> diff cd_a cd_b > cd_diff
There are 512 (569) in cd_b 768 (1012) which are not in cd_a 1096554
(1256778).
Where X (Y) - X is contents and Y is instances.

cd_diff should have the same structure as cd_a

genadyp:
So diff report actually is in the same format as ContentData?
Then it is very close to what we have in [file_utils.rb L179](https://github.com/bbfsdev/bbfs/blob/master/lib/file_utils/file_utils.rb#L179).
We can add few commented lines in the header of the report with an information similar to the command line diff or version control diff (what files were compared, etc).

bbfsdev:
Yes. That what I understand.

genadyp:
Actually we can stay with a file_utils code while:
- we add file_utils to the [gemspec](https://github.com/bbfsdev/bbfs/blob/master/content_server.gemspec#L45), then it will be installed in a system along with other executables
- return a `ContentData.to_s`. i.e. to add `output.to_s` in [file_utils.rb#L100](https://github.com/bbfsdev/bbfs/blob/master/lib/file_utils/file_utils.rb#L100)

Usage for this specific case: 
file_utils --command=minus --cd_a=master.cd --cd_b=remote.cd --dest=absents.cd

@yarondbb 
@kolmanv 
@AlexeyNemytov 
[...]

PR #255, opened 2014-03-02
Get a diff between old and new content data files

genadyp:
1.code moved from lib to bin 
2.file_utils installed in the system along with other executables 
3. output to STDOUT added
4. add parameter check for ContentData.to_file method
5.help added

Proposed solution for #246

### DIFF
--- a/bin/file_utils
+++ b/bin/file_utils
@@ -2,11 +2,119 @@
 
 # This utility executable is used to easily execute BBFS operations on files.
 
-require 'file_utils'
+require 'content_data'
 require 'log'
 require 'params'
 
+#############################################
+# Definitions
+#############################################
+Params.string 'command', nil ,'path'
+Params.string 'dest', nil ,'destination path'
+Params.string 'cd_a', nil ,'a path'
+Params.string 'cd_b', nil ,'b path'
+
+HELP_MSG = <<EOF
+This is an utility functionality for BBFS,
+such as algebra of sets operations on ContentData files.
+Usage:
+  To get merge of two ContentData files,
+  i.e. contents and appropreate instances that exist in at least one ContentData file:
+    file_utils --command=merge --cd_a=<content_data_file1> --cd_b=<content_data_file_2> [--dest=<result_content_data_file>]
+  To get intersection of two ContentData files,
+  i.e. contents and appropreate instances that exist in the both ContentData files:
+    file_utils --command=intersect --cd_a=<content_data_file1> --cd_b=<content_data_file_2> [--dest=<result_content_data_file>]
+  To get content_data_2 minus content_data_1,
+  i.e. contents and appropreate instances that exist in content_data_file2 and absent in content_data_file1:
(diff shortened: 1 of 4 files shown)


bbfsdev:
Review done.

bin/file_utils
```diff
@@ -2,11 +2,119 @@
...
 
-require 'file_utils'
+require 'content_data'
 require 'log'
 require 'params'
 
+#############################################
+# Definitions
+#############################################
+Params.string 'command', nil ,'path'
```
bbfsdev: Why command is a path?

bin/file_utils
```diff
@@ -2,11 +2,119 @@
...
+Usage:
+  To get merge of two ContentData files,
+  i.e. contents and appropreate instances that exist in at least one ContentData file:
+    file_utils --command=merge --cd_a=<content_data_file1> --cd_b=<content_data_file_2> [--dest=<result_content_data_file>]
+  To get intersection of two ContentData files,
+  i.e. contents and appropreate instances that exist in the both ContentData files:
+    file_utils --command=intersect --cd_a=<content_data_file1> --cd_b=<content_data_file_2> [--dest=<result_content_data_file>]
+  To get content_data_2 minus content_data_1,
+  i.e. contents and appropreate instances that exist in content_data_file2 and absent in content_data_file1:
+    file_utils --command=minus --cd_a=<content_data_file1> --cd_b=<content_data_file_2> [--dest=<result_content_data_file>]
```
bbfsdev, line 29: consider adding diff command (not symmetric)
it is intersect, then minus.

bin/file_utils
```diff
@@ -2,11 +2,119 @@
...
 Log.init
 
-FileUtils::FileUtils.run
+#############################################
+# Methods
+#############################################
+
+# Algebra of sets operations (merge, intersect, minus) on ContentData files
+def content_data_command()
+  if Params['cd_a'].nil?
```
bbfsdev: Isn't Params have a required field? Please use it and remove those lines.

bin/file_utils
```diff
@@ -2,11 +2,119 @@
...
+  rescue Exception => e
+    err_msg = "Error loading content data cd_a = %s: %s" % [Params['cd_a'], e.message]
+    puts err_msg
+    Log.error(err_msg)
+    Log.flush
+    return
+  end
+
+  if Params['cd_b'].nil?
+    err_msg = "--cd_b is not set"
```
bbfsdev: This code repeats itself.

bbfsdev:
Finished review

bin/file_utils
genadyp: fixed

bin/file_utils
genadyp, line 29: Is it make sense? need a discussion.

bin/file_utils
genadyp: we check here that a param was set

bin/file_utils
genadyp: I know :) Isn't more readable in this case?

PR #256, opened 2014-03-12
Issue 249 refactor monitoring

bbfsdev:
Please review the code


lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
       end
     end
 
+    ################ All monitoring helper methods #################
+    
+    # Checks that the blobed path is valid.
```
genadyp: **g**lobed path ?

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+    def globed_path_not_valid(globed_path)
```
genadyp: suggested to rename to **is_**globed_path_not_valid**?**
looks more readable and implies that return value is boolean

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+      # UTF-8 - keep only files with names in
+      return true if @non_utf8_paths[globed_path]
+      check_utf_8_encoding_file = globed_path.clone
+      unless check_utf_8_encoding_file.force_encoding("UTF-8").valid_encoding?
+        Log.warning("Non UTF-8 file name '#{check_utf_8_encoding_file}', skipping.")
+        @non_utf8_paths[globed_path]=true
+        check_utf_8_encoding_file=nil
```
genadyp: Is this for GC purposes?
If so suggested to add a comment to prevent removing this line in the future

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+    ################ All monitoring helper methods #################
+    
+    # Checks that the blobed path is valid.
+    def globed_path_not_valid(globed_path)
+      # UTF-8 - keep only files with names in
+      return true if @non_utf8_paths[globed_path]
+      check_utf_8_encoding_file = globed_path.clone
+      unless check_utf_8_encoding_file.force_encoding("UTF-8").valid_encoding?
+        Log.warning("Non UTF-8 file name '#{check_utf_8_encoding_file}', skipping.")
+        @non_utf8_paths[globed_path]=true
```
genadyp: empty spaces around `=` are absent

lib/file_monitoring/monitor_path.rb
genadyp: empty spaces around `=` are absent

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+        check_utf_8_encoding_file=nil
+        return true
+      end
+
+      return false
```
genadyp: `true` instead of `return true`
then looks more 'ruby'

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+        write_to_log("CHANGED file: " + globed_path)
+        # remove file with changed checksum. File will be added once indexed
+        $local_content_data_lock.synchronize{
+          $local_content_data.remove_instance(Params['local_server_name'], globed_path)
+        }
+      else  # case child_stat did not change
+        # ---------- SAME STATUS
+        # File status is the same
+        if child_stat.state != FileStatEnum::STABLE
+          child_stat.state = FileStatEnum::UNCHANGED
```
genadyp: suggested to move this line under the else in the line 356,
cause in the case of STABLE there is no need to set status to UNCHANGED and few lines below to set it to STABLE

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+      end
+    end
+
+    def new_manual_mode_file(globed_path, globed_path_stat, file_attr_to_checksum)
+      # --------------------- MANUAL MODE
+      # check if file name and attributes exist in global file attr map
+      file_attr_key = [File.basename(globed_path), globed_path_stat.size, globed_path_stat.mtime.to_i]
+      file_ident_info = file_attr_to_checksum[file_attr_key]
+      # If not found (real new file) or found but not unique then file needs indexing. skip in manual mode.
+      if file_ident_info and file_ident_info.unique
```
genadyp: suggested `&&` instead of `and`. as far as i remember you sent once an article about it.

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+            child_stat.state = FileStatEnum::STABLE
+            write_to_log("STABLE file: " + globed_path)
+          else
+            write_to_log("UNCHANGED file: " + globed_path)
+          end
+        end
+      end
+    end
+
+    def new_manual_mode_file(globed_path, globed_path_stat, file_attr_to_checksum)
```
genadyp: What is `file_attr_to_checksum` here? Lack of explanation :(

genadyp: Not clear what this method purpose

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+        child_stat = DirStat.new(globed_path)
+        add_dir(child_stat)
+        write_to_log("NEW dir: " + globed_path)
+      end
+      child_stat.marked = true
+      # recursive call for dirs
+      child_stat.monitor(file_attr_to_checksum)
+    end
+
+    def add_found_symlinks(globed_path, found_symlinks)
```
genadyp, line 414: Do we need to handle symlinks that point out of the monitored directories?

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+      if File.symlink?(globed_path)
+        pointed_file_name = File.readlink(globed_path)
+        found_symlinks[globed_path] = pointed_file_name
+        # add to content data
+        $local_content_data_lock.synchronize{
+          $local_content_data.add_symlink(Params['local_server_name'], globed_path, pointed_file_name)
+        }
+        return true
+      end
+      return false
```
genadyp: suggested `false` instead of `return false`
need to check what is a ruby code style in such a cases.

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+    end
+
+    def remove_not_found_symlinks(found_symlinks)
+      # check if any symlink was removed and update current symlinks map
+      symlinks_enum = @symlinks.each_key
+      loop {
+        symlink_key = symlinks_enum.next rescue break
+        if !found_symlinks.has_key?(symlink_key)
```
genadyp: suggested `unless`

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+          # symlink was removed. remove from content data
+          $local_content_data_lock.synchronize{
+            $local_content_data.remove_symlink(Params['local_server_name'], symlink_key)
+          }
+        end
+      }
+      @symlinks = found_symlinks
```
genadyp, line 437: From the GC point of view - is it better to set to nil and then to set to the correct value.
There was something that Yaron explained.

lib/file_monitoring/monitor_path.rb
genadyp, line 414: Need a comment about method return value.
From the method name it is not obvious that this method returns boolean.

lib/file_monitoring/monitor_path.rb
```diff
@@ -313,6 +314,127 @@ def removed_unmarked_paths
...
+    end
+
     # Recursively, read files and dirs from file system (using Glob)
```
genadyp: Whether algorithm description should be revised?

lib/file_monitoring/monitor_path.rb
```diff
@@ -364,112 +468,26 @@ def monitor(file_attr_to_checksum=nil)
...
-        symlink_key = symlinks_enum.next rescue break
-        if !found_symlinks.has_key?(symlink_key)
-          # symlink was removed. remove from content data
-          $local_content_data_lock.synchronize{
-            $local_content_data.remove_symlink(Params['local_server_name'], symlink_key)
-          }
-        end
-      }
-      @symlinks = found_symlinks
+      remove_not_found_symlinks(found_symlinks)
```
genadyp, line 497: Where we handles deleted files?
Should they be removed from $local_content_data ?

spec/file_monitoring/monitor_path_spec.rb
```diff
@@ -26,12 +24,112 @@ module Spec
         @dir_stat = DirStat.new('temp_path')
       end
 
+      # Test will:
```
genadyp, line 27: Could you map please this points whit  the test body itself.
For example
1) Create a monitoring directory.
......Code that creates a monitoring dir....
.................................................................
 2) Setup dummy files for glob.
....Code that sets dummy files....
...................................................

Then it looks more obvious

spec/file_monitoring/monitor_path_spec.rb
```diff
@@ -26,12 +24,112 @@ module Spec
...
+
+        print "##### Monitor 5 times #####\n"
+        # Perform monitoring
+        X.times do
+          @dir_stat.monitor
+          @dir_stat.removed_unmarked_paths
+          @dir_stat.index
+        end
+        files = @dir_stat.instance_variable_get(:@files)
+        print files
```
genadyp: Do we need to print during tests? It adds a mess to the screen when tests are executed.

spec/file_monitoring/monitor_path_spec.rb
```diff
@@ -26,12 +24,112 @@ module Spec
...
+        @dir_stat.monitor
+        @dir_stat.removed_unmarked_paths
+        @dir_stat.index
+
+        # check that files were added to Dir
+        files = @dir_stat.instance_variable_get(:@files)
+        print files
+        print "\n"
+        files.has_key?('temp_path/const_file').should be(true)
+        files.has_key?('temp_path/will_be_deleted_file').should be(false)
```
genadyp, line 95: Why it is false?

lib/file_monitoring/monitor_path.rb
yarondbb, line 437: In the end this was not part of the rules which made a difference.
GC major rules were: 
Loop using enumerator. 
Break down big data to small chunks - Use nested methods to handle the small chunks and GC in upper level of the stack.
Minimize the amount of handled data, especially sync data hanlding and avoid parallel big data manipulations.

lib/file_monitoring/monitor_path.rb
yarondbb: check_utf_8_encoding_file=nil
  In the end this was not part of the rules which made a difference.
GC major rules were: 
Loop using enumerator. 
Break down big data to small chunks - Use nested methods to handle the small chunks and GC in upper level of the stack.
Minimize the amount of handled data, especially sync data hanlding and avoid parallel big data manipulations.

lib/file_monitoring/monitor_path.rb
yarondbb, line 414: I asked Shiray regarding this and Yes we need to keep the target as it is (this is oportunistic approach)
When we recreate the symlink we will use this target. if the target will be valid we gain.
If not the symlink could be created but will not be usefull, but this is on the user responsibility since he did not put the target in a safe location.

lib/file_monitoring/monitor_path.rb
yarondbb, line 497: You mean symlinks or regular files?
For symlinks we remove files which were found at some point in past and now missing the method in line 490.
For regular files we remove it using the "mark" algorithm -  after monitoring method marks all found files & dirs, then method:removed_unmarked_paths will remove all dirs and files which are not marked.
Symlink are not part of the mark algorithm since they are very simple map.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev: Added TODO comment.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done. Nice :) now we see that splitting code is good.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev, line 414: Nice. Done.

lib/file_monitoring/monitor_path.rb
bbfsdev: Removed.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev, line 437: Skipping.

lib/file_monitoring/monitor_path.rb
bbfsdev: Done.

lib/file_monitoring/monitor_path.rb
bbfsdev, line 497: Why cannot we use the same approach with files/dirs?

spec/file_monitoring/monitor_path_spec.rb
bbfsdev, line 27: Done.

spec/file_monitoring/monitor_path_spec.rb
bbfsdev: Removed.

spec/file_monitoring/monitor_path_spec.rb
bbfsdev, line 95: This file is moved under other directory it is true under 
`a_dir_files.has_key?('temp_path/a_dir/will_be_moved_file').should be(true)`

bbfsdev:
Done!!!

PR #259, opened 2014-03-21
Fix content data format

bbfsdev:
Added read write using CSV
Added read write using gzip (compression). Now in order to watch the content_data file use zcat.
Fixed test to include symlinks. Fixed symlinks wrong reading (missing \n)
Added conversion command from old to new format
example:
bin/fix_content_data ~/.bbfs/var/content.data /tmp/fixed_content.data

### DIFF
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -2,6 +2,7 @@ require 'csv'
 require 'content_server/server'
 require 'log'
 require 'params'
+require 'zlib'
 
 module ContentData
   Params.string('instance_check_level', 'shallow', 'Defines check level. Supported levels are: ' \
@@ -326,14 +327,14 @@ module ContentData
     def to_file(filename)
       content_data_dir = File.dirname(filename)
       FileUtils.makedirs(content_data_dir) unless File.directory?(content_data_dir)
-      CSV.open(filename, "wb") do |csv|
-        csv << [@instances_info.length]
+      Zlib::GzipWriter.open(filename) do |gz| 
+        gz.write [@instances_info.length].to_csv
         each_instance { |checksum, size, content_mod_time, instance_mod_time, server, path, inst_index_time|
-          csv << [checksum, size, server, path, instance_mod_time, inst_index_time]
+          gz.write [checksum, size, server, path, instance_mod_time, inst_index_time].to_csv
         }
-        csv << [@symlinks_info.length]
+        gz.write [@symlinks_info.length].to_csv
         each_symlink { |file, path, target|
-          csv << [file, path, target]
+          gz.write [file, path, target].to_csv
         }
       end
     end
@@ -345,45 +346,47 @@ module ContentData
 
       number_of_instances = nil
       number_of_symlinks = nil
-      CSV.foreach(filename) do |row|
-        if number_of_instances == nil
-          # get number of instances
-          number_of_instances = row[0]
-          unless (number_of_instances and number_of_instances.match(/^[\d]+$/))  # check that line is of Number format
-            raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
-                  "number of instances should be a Number. We got:#{number_of_instances}")
+      Zlib::GzipReader.open(filename) { |gz|
+        gz.each_line do |line|
+          row = line.parse_csv
+          if number_of_instances == nil
+            # get number of instances
+            number_of_instances = row[0]
+            unless (number_of_instances and number_of_instances.match(/^[\d]+$/))  # check that line is of Number format
+              raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                    "number of instances should be a Number. We got:#{number_of_instances}")
+            end
+            number_of_instances = number_of_instances.to_i
+          elsif number_of_instances > 0
+            if (6 != row.length)
+              raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                        "Expected to read 6 fields ('<' separated) but got #{row.length}.\nLine:#{instance_line}")
+            end
+            add_instance(row[0],        #checksum
+                         row[1].to_i,   # size
+                         row[2],        # server
+                         row[3],        # path
+                         row[4].to_i,   # mod time
+                         row[5].to_i)   # index time
+            number_of_instances -= 1
+          elsif number_of_symlinks == nil
+            # get number of symlinks
+            number_of_symlinks = row[0]
+            unless (number_of_symlinks and number_of_symlinks.match(/^[\d]+$/))  # check that line is of Number format
+              raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                    "number of symlinks should be a Number. We got:#{number_of_symlinks}")
+            end
+            number_of_symlinks = number_of_symlinks.to_i
+          elsif number_of_symlinks > 0
+            if (3 != row.length)
+              raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                    "Expected to read 3 fields ('<' separated) but got #{row.length}.\nLine:#{symlinks_line}")
+            end
+            @symlinks_info[[row[0], row[1]]] = row[2]
+            number_of_symlinks -= 1 
           end
-          number_of_instances = number_of_instances.to_i
-        elsif number_of_instances > 0
-          if (6 != row.length)
-            raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
-                      "Expected to read 6 fields ('<' separated) but got #{row.length}.\nLine:#{instance_line}")
-          end
-          add_instance(row[0],        #checksum
-                       row[1].to_i,   # size
-                       row[2],        # server
-                       row[3],        # path
-                       row[4].to_i,   # mod time
-                       row[5].to_i)   # index time
-          number_of_instances -= 1
-        elsif number_of_symlinks == nil
-          # get number of symlinks
-          number_of_symlinks = row[0]
-          unless (number_of_symlinks and number_of_symlinks.match(/^[\d]+$/))  # check that line is of Number format
-            raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
-                  "number of symlinks should be a Number. We got:#{number_of_symlinks}")
-          end
-          number_of_symlinks = number_of_symlinks.to_i
-        elsif number_of_symlinks > 0
-          if (3 != row.length)
-            raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
-                  "Expected to read 3 fields ('<' separated) but got #{row.length}.\nLine:#{symlinks_line}")
-          end
-
-          @symlinks_info[[row[0], row[1]]] = row[2]
-          number_of_symlinks -= 1 
         end
-      end
+      }
     end
 
     ############## DEPRECATED: Old deprecated from/to file methods still needed for migration purposes
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -1,3 +1,4 @@
+require 'csv'
 require 'content_server/server'
 require 'log'
 require 'params'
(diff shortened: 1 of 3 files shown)


genadyp:
Need to update Gemfile and gemspec with new dependencies.

lib/content_data/content_data.rb
```diff
@@ -306,6 +308,15 @@ def to_s
       return_str << contents_str
       return_str << "%d\n" % [@instances_info.length]
       return_str << instances_str
+
+
+      symlinks_str = ""
+      symlinks_info_enum = @symlinks_info.each_key
```
genadyp: looks that symlinks_info_enum is unused.

lib/content_data/content_data.rb
```diff
@@ -306,6 +308,15 @@ def to_s
...
+      each_symlink { |file, path, target|
+        symlinks_str << "%s,%s,%s\n" % [file, path, target]
+      }
+      return_str << symlinks_str
```
genadyp, line 318: Number of symlinks should be added before symlinks data.
At least this is a convention with contents and instances.

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
     def to_file(filename)
       content_data_dir = File.dirname(filename)
       FileUtils.makedirs(content_data_dir) unless File.directory?(content_data_dir)
+      Zlib::GzipWriter.open(filename) do |gz|
```
genadyp, line 327: Do not see that contents also written into the file.
According to this code only instances and symlinks are written.

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+
+      number_of_instances = nil
+      number_of_symlinks = nil
+      Zlib::GzipReader.open(filename) { |gz|
+        gz.each_line do |line|
+          row = line.parse_csv
+          if number_of_instances == nil
+            # get number of instances
+            number_of_instances = row[0]
+            unless (number_of_instances and number_of_instances.match(/^[\d]+$/))  # check that line is of Number format
```
genadyp: Is an intention for:
unless (number_of_instances.**nil?**
What are incorrect values of row[0] can be?

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+      end
+
+      number_of_instances = nil
+      number_of_symlinks = nil
+      Zlib::GzipReader.open(filename) { |gz|
+        gz.each_line do |line|
+          row = line.parse_csv
+          if number_of_instances == nil
+            # get number of instances
+            number_of_instances = row[0]
```
genadyp: may be it is better to change it to try convert to integer and then to check that such a conversion succeed?

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+          gz.write [checksum, size, server, path, instance_mod_time, inst_index_time].to_csv
+        }
+        gz.write [@symlinks_info.length].to_csv
+        each_symlink { |file, path, target|
+          gz.write [file, path, target].to_csv
+        }
+      end
+    end
+
+    def from_file(filename)
```
genadyp, line 341: Need to add a comment that this method raises exeption and then this method should be run in begin..rescue block.
Otherwise it is hidden from the user of the method.

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+              raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                        "Expected to read 6 fields ('<' separated) but got #{row.length}.\nLine:#{instance_line}")
+            end
+            add_instance(row[0],        #checksum
+                         row[1].to_i,   # size
+                         row[2],        # server
+                         row[3],        # path
+                         row[4].to_i,   # mod time
+                         row[5].to_i)   # index time
+            number_of_instances -= 1
```
genadyp, line 370: What happens when number_of_instances < 0.
Should we inform a user somehow?

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+
+    def from_file(filename)
+      unless File.exists? filename
+        raise ArgumentError.new "No such a file #{filename}"
+      end
+
+      number_of_instances = nil
+      number_of_symlinks = nil
+      Zlib::GzipReader.open(filename) { |gz|
+        gz.each_line do |line|
```
genadyp, line 349: No need to read contents?
Do we add them dynamically when adding instances?

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+            add_instance(row[0],        #checksum
+                         row[1].to_i,   # size
+                         row[2],        # server
+                         row[3],        # path
+                         row[4].to_i,   # mod time
+                         row[5].to_i)   # index time
+            number_of_instances -= 1
+          elsif number_of_symlinks == nil
+            # get number of symlinks
+            number_of_symlinks = row[0]
```
genadyp: same as for instances

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+                    "number of symlinks should be a Number. We got:#{number_of_symlinks}")
+            end
+            number_of_symlinks = number_of_symlinks.to_i
+          elsif number_of_symlinks > 0
+            if (3 != row.length)
+              raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
+                    "Expected to read 3 fields ('<' separated) but got #{row.length}.\nLine:#{symlinks_line}")
+            end
+            @symlinks_info[[row[0], row[1]]] = row[2]
+            number_of_symlinks -= 1
```
genadyp, line 385: same as for instances

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
...
+                    "Expected to read 3 fields ('<' separated) but got #{row.length}.\nLine:#{symlinks_line}")
+            end
+            @symlinks_info[[row[0], row[1]]] = row[2]
+            number_of_symlinks -= 1 
+          end
+        end
+      }
+    end
+
+    ############## DEPRECATED: Old deprecated from/to file methods still needed for migration purposes
```
genadyp, line 391: Suggested to us [@deprecated tag](http://rubydoc.info/gems/yard/file/docs/Tags.md#deprecated) instead

genadyp, line 391: Do we need a file write method for the migration purposes?
May be it is sufficient only read method?

spec/content_data/content_data_spec.rb
```diff
@@ -134,6 +138,32 @@
...
+    file_moc_object = StringIO.new
+    file_moc_object.write(content_data.to_s)
+    test_file = Tempfile.new('content_data_spec.test')
+    content_data.to_file_old(test_file)
+    content_data_2 = ContentData::ContentData.new
+    content_data_2.from_file_old(test_file)
+    (content_data == content_data_2).should == true
+  end
+
+  it 'test merge' do
```
genadyp, line 161: What is the purpose of this method?
Also its name is the same as the next test method.

lib/content_data/content_data.rb
genadyp, line 391: Consider to move this deprecated code to the migration tool itself (using class instance mixins for example), then our ContentData class will be clean

lib/content_data/content_data.rb
```diff
@@ -306,6 +308,15 @@ def to_s
       return_str << contents_str
       return_str << "%d\n" % [@instances_info.length]
       return_str << instances_str
+
+
+      symlinks_str = ""
+      symlinks_info_enum = @symlinks_info.each_key
+      each_symlink { |file, path, target|
```
yarondbb: parameters list should be: { | server, path, target |

lib/content_data/content_data.rb
```diff
@@ -306,6 +308,15 @@ def to_s
...
+        symlinks_str << "%s,%s,%s\n" % [file, path, target]
+      }
+      return_str << symlinks_str
+
       return_str
     end
```
yarondbb, line 321: In general we should NOT use the to_s string but only for testings since it has very bad memory performence.
I guess that in Ruby2.1 the performence is improved (we need to check)

genadyp, line 321: Need an appropriate comment, otherwise we can fall into using it

lib/content_data/content_data.rb
```diff
@@ -316,6 +327,76 @@ def to_s
     def to_file(filename)
```
yarondbb, line 324: Since this method do not use chunks any more, the description (above the definition) should be changed

yarondbb, line 324: Also, I am not sure how would the memory performence be since chunks where used for optimizing the GC.
Without chunks, the system GC will not free objects do the the method stack contex

lib/content_data/content_data.rb
yarondbb, line 341: We have a begin rescue at the bin level. I think this is OK. The user will get the error message if something is wrong. Ususaly since the application writes and file then error in read would occur at starting the application (if user changed manually the file format for example). It would not occur at middle of execution

lib/content_data/content_data.rb
yarondbb, line 349: Adding instance will also add a content (for the first instance of that content)
Question is regarding user format. The user is expecting to see number of contents and contents right ?

lib/content_data/content_data.rb
yarondbb, line 370: Other that the issue that contents are not written\read, the algorithm is OK (number_of_instances is started at NIL then get the actauall number of instances and become as a counter to zero. When reach 0 the if else edvances..)
Problem is the it is a bit not human readable..

lib/content_data/content_data.rb
bbfsdev: Done.

lib/content_data/content_data.rb
bbfsdev, line 318: This is not the export/import function, this is helper to_s function...

lib/content_data/content_data.rb
bbfsdev, line 327: Correct, writing contents is redundant.

lib/content_data/content_data.rb
bbfsdev: Done. Numbers are ok, we are checking regex.

lib/content_data/content_data.rb
bbfsdev: Done.

lib/content_data/content_data.rb
bbfsdev, line 341: Added a comment.

lib/content_data/content_data.rb
bbfsdev, line 370: What do you mean not human readable? Please tell me so I can improve. Don't leave me with such comments.

bbfsdev, line 370: I think checking for integer is good enough, negatives are really not an option. Where is the limit of things we should check?

lib/content_data/content_data.rb
bbfsdev, line 349: Correct.
Don't think so, user don't care much about the hashes.

lib/content_data/content_data.rb
bbfsdev: Done.

lib/content_data/content_data.rb
bbfsdev, line 385: Same answer.

lib/content_data/content_data.rb
bbfsdev, line 391: We need the write method to check with tests that all works well. This is very comfortable and safe.
Then the tests will fail. I will leave the code here. We will delete it in next version.

spec/content_data/content_data_spec.rb
bbfsdev, line 161: Removed.

lib/content_data/content_data.rb
bbfsdev: Nice catch. Thanks.

lib/content_data/content_data.rb
bbfsdev, line 321: Done.

lib/content_data/content_data.rb
bbfsdev, line 324: Edited comments.

bbfsdev:
Done, may I merge?

Issue #260, opened 2014-03-22
ContentData file utils performance issue


genadyp:
**Issue description:**

> 2)  "file_utils" is able to run with "content_server" version 1.5.0
> But It works very slow. I did not get a result during 15 minutes.
> So I have terminated it. (Shirai)
#### TODO
- [x] Rake task for performance (see Rspec tag usage). `spec` task should be updated not to run performance tests. [_Rake perf and Rake all_specs are added, Rake spec run w/o performance tests_]  
- [x] Memory performance checks are too inaccurate. Consider replace/remove.
  [_Fixed: `ps` returns memory usage in KB and we assume bytes._]
- [x] GoogleHashes implementation [ #262 ]
- [ ] Update BBFS specifications, if we have any
- [ ] ContentData cloning performance improvement. [_Worth to wait for the results with GoogleHashes_]
- [ ] Ruby 2.1.1 as a project requirement

genadyp:
Working branch: [issue260_diff_performance](https://github.com/bbfsdev/bbfs/tree/issue260_diff_performance)

genadyp:
**Content Data Performance Test**
1. Object initialization  
   1. Init one object  
      1. 350000 instances in less then 600 seconds  
      2. clone from exist object of 350000 in less thes 600 seconds  
   2. Init more then one object  
      1. two object each of 350000 instances in less then 1200 seconds
2. File operations  
   1. save/load on 350000 instances in less then 600 seconds  
3. Set operations
   1. minus of two objects with 350000 instances each
      1. finish in less then 600 seconds

**Results:**

> ContentData init for 350000 instancesfinished in 7 seconds
> **Clone from exist object of 350000 instances finished in 382 seconds**
> Init of 2 ContentData of 350000 instances each finished in 34 seconds
> **To file for 350000 finished in 325 seconds**
> From file for 350000 finished in 34 seconds
> **ContentData minus for 350000 finished in 536 seconds**

**Problems:**
1. Clone operation
   1. minus operation depends on clone
2. To file
   1. takes x10 times then from file
3. ~~Looks like more objects we initialize - it takes more time per initialization~~  
     Init of 3 ContentData of 350000 instances each finished in 54 seconds
     while init of 2 objects finished in 34 seconds.
     Looks like there is a jump only for more then 1 object and there is no geometric progression.

kolmanv:
1) Why minus depends on clone? This might not be necessary.
[...]

PR #263, opened 2014-04-07
Issue260 diff performance

genadyp:
connected to #260 

---

Tests ContentData specifications:
basic ContentData methods on 350_000 instances
- finish in less then within 5 minutes
- consume less then 250 MB

NOTE the results are not exact cause they do not run in a clean environment and influenced from
monitoring/testing code, but they give a good approximation
Supposition: monitoring code penalty is insignificant against time/memory usage of the tested code

---

Was found that ConntentData cloning is extremely expensive operation

> Init one object 350000 instances in less then 300 seconds: finished in **5** seconds
> Init one object clone of 350000 in less then 300 seconds: finished in **59** seconds
> Init more then one object two object of 350000 instances each in less then 600 seconds: finished in 9 seconds
> Init more then one object three object of 350000 instances each in less then 900 seconds: finished in 15 seconds
> Init more then one object three object of 350000 instances each in less then 900 seconds: finished in 15 seconds

then minus of two ContentData objects was re-written without using cloning 

---

Was found that performance of Ruby2.1.1 is much better than Ruby1.9.3

> $ rvm current
> **ruby-1.9.3-p327**
> $ rspec -fd spec/content_data/content_data_performance_spec.rb
> .....
> minus of two objects with 350000 instances each finish in less then 300 seconds: **do not finished in 300 seconds**
> 
> $ rvm current
> ruby-2.1.1
> $ rspec -fd spec/content_data/content_data_performance_spec.rb
> ......
> minus of two objects with 350000 instances each finish in less then 300 seconds: **finished in 25 seconds**

**Then it is highly recommended to move to Ruby2.1.1**
With Ruby1.9.3 we still remain above 10 minutes for minus operation

### DIFF
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -767,7 +767,8 @@ module ContentData
     return ContentData.new(b) if a.nil?
     c = ContentData.new
     b.each_instance do |checksum, size, _, instance_mtime, server, path, index_time|
-      unless (a.instance_exists(path, server))
+      #unless (a.instance_exists(path, server))
+      unless (a.content_exists(checksum))
               c.add_instance(checksum,
                              size,
                              server,
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -31,6 +31,8 @@ module ContentData
 
     CHUNK_SIZE = 5000
 
+    # NOTE Cloning is time/memory expensive operation.
+    # It is highly recomended to avoid it.
     def initialize(other = nil)
       if other.nil?
         @contents_info = {}  # Checksum --> [size, paths-->time(instance), time(content)]
@@ -763,11 +765,17 @@ module ContentData
   def self.remove(a, b)
     return nil if b.nil?
     return ContentData.new(b) if a.nil?
-    c = ContentData.new(b)  # create new cloned content C from B
-    # remove contents of A from newly cloned content A
-    a.each_content { |checksum, size, content_mod_time|
-      c.remove_content(checksum)
-    }
+    c = ContentData.new
+    b.each_instance do |checksum, size, _, instance_mtime, server, path, index_time|
+      unless (a.instance_exists(path, server))
+              c.add_instance(checksum,
+                             size,
+                             server,
+                             path,
+                             instance_mtime,
+                             index_time)
+      end
+    end
     c
   end
 
--- a/spec/content_data/content_data_performance_spec.rb
+++ b/spec/content_data/content_data_performance_spec.rb
@@ -12,7 +12,7 @@ require_relative '../../lib/content_data/content_data.rb'
 describe 'Content Data Performance Test' do
 
   NUMBER_INSTANCES = 350_000
-  MAX_CHECKSUM = NUMBER_INSTANCES  #1000
+  MAX_CHECKSUM = NUMBER_INSTANCES
   INSTANCE_SIZE = 1000
   SERVER = "server"
   PATH = "file_"
@@ -55,7 +55,7 @@ describe 'Content Data Performance Test' do
         [build_thread, timer_thread].each { |th| th.join }
 
         is_succeeded = timer < LIMIT_TIME
-        msg = "ContentData build for #{NUMBER_INSTANCES} " +
+        msg = "ContentData init for #{NUMBER_INSTANCES} " +
           (is_succeeded ? "" : "do not ") + "finished in #{timer} seconds"
 
         # main check
@@ -91,7 +91,7 @@ describe 'Content Data Performance Test' do
         [init_thread, timer_thread].each { |th| th.join }
 
         is_succeeded = timer < LIMIT_TIME
-        msg = "ContentData build for #{NUMBER_INSTANCES} " +
+        msg = "ContentData init from exist object of #{NUMBER_INSTANCES} " +
           (is_succeeded ? "" : "do not ") + "finished in #{timer} seconds"
 
         # main check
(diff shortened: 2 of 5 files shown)


lib/content_data/content_data.rb
```diff
@@ -763,11 +765,18 @@ def self.merge_override_b(a, b)
...
     return nil if b.nil?
     return ContentData.new(b) if a.nil?
-    c = ContentData.new(b)  # create new cloned content C from B
-    # remove contents of A from newly cloned content A
-    a.each_content { |checksum, size, content_mod_time|
-      c.remove_content(checksum)
-    }
+    c = ContentData.new
+    b.each_instance do |checksum, size, _, instance_mtime, server, path, index_time|
+      #unless (a.instance_exists(path, server))
```
bbfsdev, line 770: Please remove this comment.

spec/content_data/content_data_performance_spec.rb
```diff
@@ -0,0 +1,321 @@
...
+  def get_initialized
+    initialized_cd = ContentData::ContentData.new
+    NUMBER_INSTANCES.times do |i|
+      initialized_cd.add_instance(i.to_s, INSTANCE_SIZE+i, SERVER, PATH + i.to_s, MTIME+i)
+    end
+    initialized_cd
+  end
+
+  # TODO consider to separate it to 2 derived classes: one for memory, one for time monitoring
+  class Limit
```
bbfsdev, line 62: There are native implementations for time and memory performance checks.
IIUF the only thing that is missing in the "stop when time ends".

I suggest using:
http://www.ruby-doc.org/stdlib-2.1.1/libdoc/timeout/rdoc/Timeout.html
together with:
http://www.ruby-doc.org/stdlib-2.1.1/libdoc/benchmark/rdoc/Benchmark.html

```
require 'benchmark'
require 'timeout'
status = Timeout::timeout(5) {
  Benchmark.measure { "a"*1_000_000_000 }
}
```

bbfsdev, line 62: For memory probably this one is the most popular:
https://github.com/ruby-prof/ruby-prof

spec/content_data/content_data_performance_spec.rb
```diff
@@ -0,0 +1,321 @@
...
+        true
+      rescue Errno::ESRCH
+        false
+      end
+    end
+
+    # Get memory consumed my the process
+    # @param [Integer] pid, default is current pid
+    # @return [Integer] memory usage in kilobytes
+    def Process.get_memory_usage(pid = Process.pid)
```
bbfsdev, line 145: Outsourcing memory check, nice :)

spec/content_data/content_data_performance_spec.rb
```diff
@@ -0,0 +1,321 @@
...
+    # TODO consider usage of Process::setrlimit
+    # @param [Limit] limit object
+    def call_with_memory_limit(limit)
+      call_thread = Thread.new { self.call }
+      limit_thread = limit.get_memory_limit_thread(call_thread)
+      [call_thread, limit_thread].each { |th| th.join }
+    end
+  end
+
+  module Process
```
bbfsdev, line 129: Do you extend the existing Process module or create your own? Please add comment.

spec/content_server/file_streamer_spec.rb
```diff
@@ -15,7 +15,7 @@
 Params['log_write_to_file'] = false
 Params['log_debug_level'] = 0
 Params['streaming_chunk_size'] = 5
-Params.init ARGV
+Params.init Array.new
```
bbfsdev, line 18: Why? What is the difference? What parameter are coming from ARGV? something from rake?

lib/content_data/content_data.rb
genadyp, line 770: done

spec/content_data/content_data_performance_spec.rb
genadyp, line 62: The purpose of this spec was to test a specification, i.e. in this specific case to check that running time is less than a limit and do not to check what is a running time.
If we see that we are above the limit there is no taste to continue to test the specification.
The situation was that a limit was 5 minutes and a running time is at least 15 minutes, than user is not productive 10 minutes, while he is waiting for the result (practically it rather difficult to switch from one task (checking specification) to another while waiting for the result). 
Then "stop when time ends", that is missing with 'benchmark' and 'timeout' is important.
We can consider their usage in other cases.

genadyp, line 62: I tried ruby-prof. It looks complicated and requires learning.
If we will decide to use it than there will be a need to allocate enough time to learn this library.

spec/content_data/content_data_performance_spec.rb
genadyp, line 129: As far as I understand it is a mixin, while we add a new functionality to existing objects or classes.

genadyp:
Yes, rspec is running now with [tag command line parameter](https://www.relishapp.com/rspec/rspec-core/v/2-4/docs/command-line/tag-option) that determine whether to run performance specs.

[Performance tests tagged with ':tag => true' tag.](https://github.com/bbfsdev/bbfs/blob/e5ded9dba2b1265fa64601fbf33d34fc4e28dbdf/Rakefile#L153)

spec/content_data/content_data_performance_spec.rb
bbfsdev, line 62: I dont' understand this sentence:
"Then "stop when time ends", that is missing with 'benchmark' and 'timeout' is important."

Isn't timeout stops the run after X time? Isn't this what you needed? And if benchmark finishes within 5 minutes you will get the stats.

genadyp, line 62: Agree.

PR #267, opened 2014-04-26
Add google hashes

bbfsdev:
@genadyp please review.

### DIFF
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -243,7 +243,12 @@ module ContentData
       @contents_info.each { |checksum, content_info|
         locations_to_delete = []
         content_info[1].each { |location, stats|
-          if location[0] == server and location[1].scan(dir_to_remove).size > 0
+          path_split = location[1].split(File::SEPARATOR)
+          dir_split = dir_to_remove.split(File::SEPARATOR)
+          if path_split[0] != "" || dir_split[0] != ""
+            throw ArgumentError.new "File path (%s or %s) does not start with %s" % [location[1], dir_to_remove, File::SEPARATOR]
+          end
+          if location[0] == server && dir_split.size <= path_split.size && path_split[0..dir_split.size-1] == dir_split
             locations_to_delete.push(location)
           end
           locations_to_delete.each { |location|
@@ -318,7 +323,6 @@ module ContentData
       return_str << contents.sort.join("\n")
       return_str << "\n%d\n" % [@instances_info.length]
       return_str << instances.sort.join("\n") << "\n"
-
 
       symlinks = []
       each_symlink { |server, path, target|
--- a/lib/content_data/content_data.rb
+++ b/lib/content_data/content_data.rb
@@ -3,11 +3,11 @@ require 'log'
 require 'params'
 require 'google_hash'
 
-class GoogleHashSparseRubyToRuby
-  def each_key &block
-    keys.each &block
-  end
-end
+#class GoogleHashSparseRubyToRuby
+#  def each_key &block
+#    keys.each &block
+#  end
+#end
 
 module ContentData
   Params.string('instance_check_level', 'shallow', 'Defines check level. Supported levels are: ' \
@@ -63,11 +63,6 @@ module ContentData
       @instances_info.each { |key, value|
         clone_instances_info[key] = value
       }
-      #instances_info_enum = @instances_info.each_key
-      #loop {
-      #  location = instances_info_enum.next rescue break
-      #  clone_instances_info[[location[0].clone, location[1].clone]] = @instances_info[location].clone
-      #}
       clone_instances_info
     end
 
@@ -76,27 +71,6 @@ module ContentData
       @contents_info.each { |key, value|
         clone_contents_info[key] = value
       }
-      #contents_info_enum = @contents_info.each_key
-      #loop {
-      #  checksum = contents_info_enum.next rescue break
-      #  instances = @contents_info[checksum]
-      #  size = instances[0]
-      #  content_time = instances[2]
-      #  instances_db = instances[1]
-      #  instances_db_cloned = GoogleHashSparseRubyToRuby.new
-      #  instances_db_enum = instances_db.each_key
-      #  loop {
-      #    location =  instances_db_enum.next rescue break
-      #    inst_mod_times = instances_db[location]
-      #    # we use deep clone for location since map key is using shallow clone.
-      #    # we dont want references between new content data
-      #    # and orig object. This will help the GC dispose the orig object if not used any more.
-      #    instances_db_cloned[[location[0].clone,location[1].clone]] = inst_mod_times.clone
-      #  }
-      #  clone_contents_info[checksum] = [size,
-      #                        instances_db_cloned,
-      #                        content_time]
-      #}
       clone_contents_info
     end
 
@@ -105,12 +79,6 @@ module ContentData
       @symlinks_info.each{ |key, value|
         cloned_symlinks[ket] = value
       }
-      #symlinks_info_enum = @symlinks_info.each_key
-      #cloned_symlinks = GoogleHashSparseRubyToRuby.new
-      #loop {
-      #  symlink_key = symlinks_info_enum.next rescue break
-      #  cloned_symlinks[[symlink_key[0].clone, symlink_key[0].clone]] = @symlinks_info[symlink_key].clone
-      #}
       cloned_symlinks
     end
 
@@ -120,13 +88,6 @@ module ContentData
       @contents_info.each { |checksum, content_info|
         block.call(checksum, content_info[0], content_info[2])
       }
-      #contents_enum = @contents_info.each_key
-      #loop {
-      #  checksum = contents_enum.next rescue break
-      #  content_val = @contents_info[checksum]
-      #  # provide checksum, size and content modification time to the block
-      #  block.call(checksum,content_val[0], content_val[2])
-      #}
     end
 
     # iterator over @contents_info data structure (including instances)
@@ -142,20 +103,6 @@ module ContentData
           }
         }
       }
-      #contents_enum = @contents_info.each_key
-      #loop {
-      #  checksum = contents_enum.next rescue break
-      #  content_info = @contents_info[checksum]
-      #  content_info_enum = content_info[1].each_key
-      #  loop {
-      #    location = content_info_enum.next rescue break
-      #    # provide the block with: checksum, size, content modification time,instance modification time,
-      #    #   server and path.
-      #    inst_mod_time, inst_index_time = content_info[1][location]
-      #    block.call(checksum,content_info[0], content_info[2], inst_mod_time,
-      #               location[0], location[1], inst_index_time)
-      #  }
-      #}
     end
 
     # iterator of instances over specific content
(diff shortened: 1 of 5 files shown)


lib/content_data/content_data.rb
```diff
@@ -50,51 +63,33 @@ def initialize(other = nil)
...
 
     def clone_instances_info
-      clone_instances_info = {}
-      instances_info_enum = @instances_info.each_key
-      loop {
-        location = instances_info_enum.next rescue break
-        clone_instances_info[[location[0].clone, location[1].clone]] = @instances_info[location].clone
+      new_instances_info = GoogleHashSparseRubyToRuby.new
+      @instances_info.each { |key, value|
+        new_instances_info[key] = value
```
genadyp, line 72: Should  be used a clone method as it was in Yaron implementation?

lib/content_data/content_data.rb
```diff
@@ -50,51 +63,33 @@ def initialize(other = nil)
...
-          # we use deep clone for location since map key is using shallow clone.
-          # we dont want references between new content data
-          # and orig object. This will help the GC dispose the orig object if not used any more.
-          instances_db_cloned[[location[0].clone,location[1].clone]] = inst_mod_times.clone
+      new_contents_info = GoogleHashSparseRubyToRuby.new
+      @contents_info.each { |checksum, content_info|
+        new_content_info = [content_info[0], GoogleHashSparseRubyToRuby.new, content_info[2]]
+        new_contents_info[checksum] = new_content_info
+        content_info[1].each { |location, stat|
+          new_content_info[1][location] = stat
```
genadyp: Should be used a clone method as it was in Yaron implementation?

lib/content_data/content_data.rb
```diff
@@ -50,51 +63,33 @@ def initialize(other = nil)
...
 
     def clone_symlinks_info
-      symlinks_info_enum = @symlinks_info.each_key
-      cloned_symlinks = {}
-      loop {
-        symlink_key = symlinks_info_enum.next rescue break
-        cloned_symlinks[[symlink_key[0].clone, symlink_key[0].clone]] = @symlinks_info[symlink_key].clone
+      cloned_symlinks = GoogleHashSparseRubyToRuby.new
+      @symlinks_info.each{ |key, value|
+        cloned_symlinks[ket] = value
```
genadyp: Should be used a clone method as it was in Yaron implementation?

lib/content_data/content_data.rb
```diff
@@ -115,17 +106,10 @@ def each_content(&block)
     # block is provided with: checksum, size, content modification time,
     #   instance modification time, server and file path
```
genadyp: comment should be updated: "instance index time" block parameter is missing

lib/content_data/content_data.rb
```diff
@@ -171,15 +148,16 @@ def symlinks_size()
     end
 
     def checksum_instances_size(checksum)
```
genadyp, line 152: Method name is not informative.
Consider rename or add method description.

lib/content_data/content_data.rb
bbfsdev, line 72: Value here is checksum, basic value, no need to clone it. There is a need to clone only arrays and hashes.
This code is not pretty, we really should use small classes for that.

lib/content_data/content_data.rb
bbfsdev: This code is already updated with stat.clone().
Good catch.

lib/content_data/content_data.rb
bbfsdev: Value here is target file, basic string, no need to clone.

lib/content_data/content_data.rb
bbfsdev: Done.

lib/content_data/content_data.rb
bbfsdev, line 152: Added description.

bbfsdev:
Done.

lib/content_data/content_data.rb
```diff
@@ -171,15 +148,16 @@ def symlinks_size()
...
+      return 0 unless @contents_info.key?(checksum)
       content_info = @contents_info[checksum]
-      return 0 if content_info.nil?
       content_info[1].length
     end
 
     def get_instance_mod_time(checksum, location)
+      return nil unless @contents_info.key?(checksum)
```
genadyp, line 160: to raise exception in this case?
Need a method description comment when nil returned or exception raised - it makes usage more secure..

lib/content_data/content_data.rb
```diff
@@ -171,15 +148,16 @@ def symlinks_size()
...
       content_info = @contents_info[checksum]
-      return nil if content_info.nil?
       instances = content_info[1]
+      return nil
```
genadyp: this return makes following two lines unreachable

lib/content_data/content_data.rb
```diff
@@ -171,15 +148,16 @@ def symlinks_size()
     end
 
     def checksum_instances_size(checksum)
+      return 0 unless @contents_info.key?(checksum)
       content_info = @contents_info[checksum]
-      return 0 if content_info.nil?
       content_info[1].length
     end
 
     def get_instance_mod_time(checksum, location)
```
genadyp, line 159: Need somewhere a general comment what is a time format we are working with.
It is not obvious that a return value of the method is an integer (Unix epoch time).

lib/content_data/content_data.rb
```diff
@@ -256,40 +237,63 @@ def remove_instance(server, path)
...
-      loop {
-        checksum = contents_enum.next rescue break
-        instances =  @contents_info[checksum][1]
-        instances_enum = instances.each_key
-        loop {
-          location = instances_enum.next rescue break
+      @contents_info.each { |checksum, content_info|
+        locations_to_delete = []
+        content_info[1].each { |location, stats|
           if location[0] == server and location[1].scan(dir_to_remove).size > 0
```
genadyp: Looks like scan can return an unexpected results.
For example we'd like to remove directory "/dev/tools/"
then location /home/genadyp/dev/tools/ will be deleted:

> 1.9.3-p327 :022 >   dir = "/dev/tools/"
>  => "/dev/tools/" 
> 1.9.3-p327 :023 > path = "/home/genadyp/dev/tools/irb"
>  => "/home/genadyp/dev/tools/irb" 
> 1.9.3-p327 :024 > path.scan(dir)
>  => ["/dev/tools/"] 
> 1.9.3-p327 :025 > 

Is it OK?

lib/content_data/content_data.rb
bbfsdev: This was the actual but why unify time did not work :)
Nice catch.

lib/content_data/content_data.rb
bbfsdev, line 159: Done.

lib/content_data/content_data.rb
bbfsdev: No, this is not ok, should be prefix.

lib/content_data/content_data.rb
```diff
@@ -256,40 +240,68 @@ def remove_instance(server, path)
...
-        loop {
-          location = instances_enum.next rescue break
-          if location[0] == server and location[1].scan(dir_to_remove).size > 0
-            instances.delete(location)
-            @instances_info.delete(location)
+      @contents_info.each { |checksum, content_info|
+        locations_to_delete = []
+        content_info[1].each { |location, stats|
+          path_split = location[1].split(File::SEPARATOR)
+          dir_split = dir_to_remove.split(File::SEPARATOR)
```
genadyp: looks like it enough to run split on dir_to_remove only once, suggested to move it out of the loop.

lib/content_data/content_data.rb
```diff
@@ -256,40 +240,68 @@ def remove_instance(server, path)
...
+          if path_split[0] != "" || dir_split[0] != ""
```
genadyp: looks like linux specific check and than can fall on windows.

lib/content_data/content_data.rb
```diff
@@ -256,40 +240,68 @@ def remove_instance(server, path)
...
 
       # handle symlinks
-      symlinks_enum = @symlinks_info.each_key
-      loop {
-        symlink_key = symlinks_enum.next rescue break
-        if symlink_key[0] == server and symlink_key[1].scan(dir_to_remove).size > 0
-          @symlinks_info.delete(symlink_key)
+      @symlinks_info.each { |key, _|
+        keys_to_delete = []
+        if key[0] == server and key[1].scan(dir_to_remove).size > 0
```
genadyp: looks like this scan method should be treated as above

lib/content_data/content_data.rb
```diff
@@ -256,40 +240,68 @@ def remove_instance(server, path)
...
         end
+        keys_to_delete.each { |key|
+          @symlinks_info.delete(key)
+        }
       }
     end
 
     def ==(other)
-      return nil if other.nil?  # for this case: content_data == nil
-      unique_id  == other.unique_id
```
genadyp, line 286: was found any problem with '==' in google_hash ?

lib/content_data/content_data.rb
```diff
@@ -256,40 +240,68 @@ def remove_instance(server, path)
...
+        return false if other.instances_info[location] != checksum
+      }
+      return true
     end
 
     def remove_content(checksum)
       content_info = @contents_info[checksum]
       if content_info
-        content_info[1].each_key { |location|
+        content_info[1].each { |location, stats|
```
genadyp: stats parameter is unused - can be replaced by '_' in the block parameters

lib/content_data/content_data.rb
```diff
@@ -299,25 +311,24 @@ def remove_content(checksum)
...
       }
       each_instance { |checksum, size, content_mod_time, instance_mod_time, server, path|
-        instances_str <<  "%s,%d,%s,%s,%d\n" % [checksum, size, server, path, instance_mod_time]
+        instances <<  "%s,%d,%s,%s,%d" % [checksum, size, server, path, instance_mod_time]
       }
       return_str << "%d\n" % [@contents_info.length]
-      return_str << contents_str
-      return_str << "%d\n" % [@instances_info.length]
-      return_str << instances_str
+      return_str << contents.sort.join("\n")
```
genadyp: suggest to check how many time takes sort on a large array

lib/content_data/content_data.rb
bbfsdev, line 286: No, this comparison is bad, it is based on internal unique_id which is wrong here, what if we have to identical copies then the unique_id will be different. That is not a way to do comparison.

lib/content_data/content_data.rb
bbfsdev: Done.

lib/content_data/content_data.rb
bbfsdev: Removed this check.

lib/content_data/content_data.rb
bbfsdev: Done. Added tests to symlinks - there was none!!!

lib/content_data/content_data.rb
bbfsdev: Marked as unused_stats

lib/content_data/content_data.rb
bbfsdev: This is for debug only purposes, we don't use print in other situations.
Why do we need to check how many time takes sort? Should be n*log(n).
Anyway we should not do this unless we have speed problems.

bbfsdev:
Done, please approve to merge...

spec/file_utils/time_modification_spec.rb
```diff
@@ -41,6 +41,7 @@ module Spec
         # must preced Log.init, otherwise log containing default values will be created
         Params['log_write_to_file'] = false
         Params['log_write_to_console'] = false
+        Params['log_debug_level'] = 1
```
genadyp, line 44: Do we actually need to output debug messages?

spec/file_utils/time_modification_spec.rb
```diff
@@ -127,6 +128,7 @@ module Spec
         }
 
         # checking that files were actually modified
+        # TODO(kolman): Decide whether files should be updated or not and change the implementation.
```
genadyp, line 131: Consider to open issue about this TODO

genadyp:
@kolmanv 
Review is finished.
There was few minor additional comments.
You can merge.
Thank you :)

Issue #268, opened 2014-05-26
data migration bug


genadyp:
@vshiray reported:

> I have a problem with data migration:
> [bbfs@swift1 var]$ fix_content_data backup.data fixed.data
> /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/lib/log.rb:112:in `init': undefined method`each' for nil:NilClass (NoMethodError)
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/lib/log.rb:125:in `warning'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/lib/content_data/content_data.rb:192:in `add_instance'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/lib/content_data/content_data.rb:577:in `read_old_instances_chunk'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/lib/content_data/content_data.rb:517:in `block in from_file_old'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/lib/content_data/content_data.rb:476:in `open'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/lib/content_data/content_data.rb:476:in `from_file_old'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/gems/content_server-1.7.0/bin/fix_content_data:10:in `<top (required)>'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/bin/fix_content_data:23:in `load'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/bin/fix_content_data:23:in `<main>'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/bin/ruby_executable_hooks:15:in `eval'
> - from /usr/local/rvm/gems/ruby-2.1.1@bbfs/bin/ruby_executable_hooks:15:in `<main>'

genadyp:
@vshiray 
@kolmanv 
@yarondbb 
@AlexeyNemytov 

Problems:
1.No Params.init run while Log suppose that Params were initialized.
Solution: fixed in fix_content_data
2.There is no check of an instance line that was read from old-format file.
Solution: check was added
[...]

PR #269, opened 2014-05-26
fix issue268

genadyp:
Fix for #268 

1.No Params.init run while Log suppose that Params were initialized.
Solution: fixed in fix_content_data
2.There is no check of an instance line that was read from old-format file.
Solution: check was added
3.We suppose that delimiter in the old format files is a comma, but actually, in this specific case '<' was used as a delimiter.
Solution: additional parameter, that contains fields delimiter was added to fix_content_data executable and appropriate methods of ContentData. Default value was set to comma.
4. Params.init no more demands from user an input.
Motivation: there are a number of executables and tests that use Params.init but have no ARGV array.
5. command line help was added for bin/fix_content_data

### DIFF
--- a/bin/fix_content_data
+++ b/bin/fix_content_data
@@ -1,11 +1,31 @@
 #!/usr/bin/env ruby
 
-# Get two file names, one is source content data other is destinations.
-# Converts source content data old format to new format and writes it to
-# destination file.
-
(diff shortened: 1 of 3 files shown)


bbfsdev:
Looks good to me. Please merge and create a patch. 1.7.1

PR #272, opened 2014-07-25
Optional cd zip

yarondbb:


lib/content_server/backup_server.rb
```diff
@@ -58,6 +58,15 @@ def run_backup_server
     $local_content_data = ContentData::ContentData.new
     $last_content_data_id = $local_content_data.unique_id
 
+    # check if local content data file name indicates compression (ends with '.gz')
```
bbfsdev, line 61: Please extract this to one function (using this twice from different places) in server.rb file.

lib/file_monitoring/file_monitoring.rb
```diff
@@ -210,23 +210,28 @@ def monitor_files
         end
 
         # Start monitor
-        Log.info("Start monitor path:%s ", dir_stat.path)
-        $testing_memory_log.info("Start monitor path:#{dir_stat.path}") if $testing_memory_active
+        msg = "Start monitor path:#{dir_stat.path}"
```
bbfsdev, line 213: Please extract this to helper method.

lib/content_data/content_data.rb
```diff
@@ -324,6 +324,17 @@ def to_s
 
     # Write content data to file.
     def to_file(filename)
+      if $local_content_data_compressed
```
bbfsdev, line 327: I don't like much the global vars approach, but don't have any better solutions yet.

lib/content_data/content_data.rb
```diff
@@ -361,7 +388,7 @@ def from_file(filename)
           elsif number_of_instances > 0
             if (6 != row.length)
               raise("Parse error of content data file:#{filename}  line ##{$.}\n" +
-                        "Expected to read 6 fields ('<' separated) but got #{row.length}.")
+                        "Expected to read 6 fields (',' separated) but got #{row.length}.")
```
bbfsdev, line 391: Why is this part of this change?
Did you merge with master?

lib/content_data/content_data.rb
```diff
@@ -390,6 +417,54 @@ def from_file(filename)
       }
     end
 
+    # Write content data to file.
+    # Write is using chunks (for both content chunks and instances chunks)
+    # Chunk is used to maximize GC affect. The temporary memory of each chunk is GCed.
+    # Without the chunks used in a dipper stack level, GC keeps the temporary objects as part of the stack context.
+    def to_file_not_compressed(filename)
```
bbfsdev, line 424: This is not the right way to do this. Reuse the to_file method.
When generating the writer, i.e., use ether one or other:

writer = File.open(filename, 'w')
writer = Zlib::GzipWriter.open(filename)

then use writer object...

lib/content_data/content_data.rb
```diff
@@ -390,6 +417,54 @@ def from_file(filename)
...
+        symlinks_info_enum = @symlinks_info.each_key
+        file.write("#{@symlinks_info.length}\n")
+        loop {
+          symlink_key = symlinks_info_enum.next rescue break
+          file.write("#{symlink_key[0]},#{symlink_key[1]},#{@symlinks_info[symlink_key]}\n")
+        }
+      }
+    end
+
+    def to_file_instances_chunk_no_zip(file, contents_enum, chunk_size)
```
bbfsdev, line 451: no need to duplicate those functions.

lib/content_data/content_data.rb
```diff
@@ -460,11 +535,113 @@ def to_old_file_instances_chunk(file, contents_enum, chunk_size)
       end
     end
 
+    # Loading db from file using chunks
+    # Chunk is used to maximize GC affect. The temporary memory of each chunk is GCed.
+    # Without the chunks used in a dipper stack level, GC keeps the temporary objects as part of the stack context.
+    # @param filename [String] filename of the file containing ContentData in obsolete format
+    # @param delimiter [Character] fields separator character, default is ','
+    def from_file_not_compressed(filename, delimiter=',')
```
bbfsdev, line 543: Same thing here use reader object:
reader = Zlib::GzipReader.open(filename)
reader = File.open(filename)

lib/content_server/content_server.rb
```diff
@@ -47,6 +47,15 @@ def run_content_server
     $local_content_data = ContentData::ContentData.new
     $last_content_data_id = $local_content_data.unique_id
 
+    # check if local content data file name indicates compression (ends with '.gz')
```
bbfsdev, line 50: Use the extracted method from above.

bbfsdev:
Start the fixes with two tests:
1) Take fake content_data
2) Store it to .gz file, store it to regular file.
3) Load the .gz file and check it is compressed.
4) Load the regular file and check it is not compressed.

lib/content_data/content_data.rb
yarondbb, line 391: Yes the branch was from latest 'master'
I fixed this since actually we expect ',' char.
I am not clear with how we wrtie\read files with ',' in the file name

lib/content_data/content_data.rb
yarondbb, line 451: which functions are duplicated ?

lib/content_data/content_data.rb
yarondbb, line 543: We need to talk I dont follow

yarondbb:
Closed without merging. This feature is moved to other branch.
This branch will be deleted

